PR #36, opened 2019-04-22
Add code coverage tests

caique-lima:
### Status
**READY**

### Background context
Keep tests covering all our code is easier if we have coverage tests

Add configuration to local coverage tests, and add command to send reports to codecov

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -2,3 +2,4 @@ pytest>=4.2.1,<5
 pytest-cov>=2.6.1,<3
 pytest-xdist>=1.26.1,<2
 mypy>=0.670,<1
+codecov>=2.0,<3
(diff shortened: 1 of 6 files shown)


requirements_test.txt
```diff
@@ -2,3 +2,4 @@ pytest>=4.2.1,<5
 pytest-cov>=2.6.1,<3
 pytest-xdist>=1.26.1,<2
 mypy>=0.670,<1
+codecov~=2.0
```
darlansf1: We should probably agree to a single standard and either use `>=,<` or `~=` everywhere. I'm okay with both. I think the only disadvantage of tilde notation is that it's not possible to do something like `lib>=0.1.2,<0.1.5`

caique-lima: 🆗 I will go with `>=,<` to allow us use `lib>=0.1.2,<0.1.5`

PR #46, opened 2019-04-24
Git clone simplified and parentheses removed

Sphinxs:
### Status
**READY**

### Background context
The motivation was a simplicity reading of the 'get started'.

The command `git clone -b master https://github.com/nubank/fklearn.git --depth=1`, clone the latest version of the branch master. The history is truncated to the specified number of commits (depth).

The '$'s where added to the clone 'code' to specify that the command should be typed in the terminal. The [RST](https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst) language don't specify a code syntax like '``` sh ```'.

Some parentheses were removed because its verbose usage. The Pandas Dataframe was one example, it could be explicity in both cases: parameters and function return, reference to the [training](https://github.com/nubank/fklearn/blob/d125c3cdfeef06c6b9cb55dd013ddbabefd0d12b/src/fklearn/training/pipeline.py#L91) function.

From line 15 to 43 of **docs/source/getting_started.rst**

### What's not done?
Update the README.md to reflect the clone method.


prcastro:
Please update README as well, would be nice to keep things consistent :)

caique-lima:
> Nice! @caique-lima do you have any opinion about the `depth=1` part? In my understanding there are very remote chances of getting into trouble when trying to merge stuff when using this shallow clone, but this should never be a problem if you're only trying to install fklearn. I'm OK with keeping that.

In order to install the package I'm ok with that

Sphinxs:
@prcastro @caique-lima , done!

PR #49, opened 2019-04-25
validator accepts predict_oof as argument

pvcrossi:
### Status
**READY**

### Background context
`validator` was the only function in `validator.py` to not get `predict_oof` as an argument, even though it was already listed as an argument in the function's docstring.

This PR adds `predict_oof` as an optional arg to the function header and passes it as an argument to `validator_iterator` inside the function body.


pvcrossi:
> LGTM, bump version and update changelog

Done

PR #58, opened 2019-04-30
Fix issue with encoded columns in pipeline

pvcrossi:
### STATUS
**READY**

### Background context
Since `features` must be defined when creating a pipeline, new features created by the one hot encoder do not reach consequent learners.
But these feature names have a specific syntax, e.g. `col_name + "==" + val_name`, which allows them to be found from within the learners.

Features created by the one hot encoder now have a different name syntax, `fklearn_feat__col==val`, which can be better generalized in other learners.

In all classification/regression/unsupervised learners, with [new optional parameter]
`encode_extra_cols = True`, the list of features is expanded to include all other features which have names like `fklearn_feat__col==val` and `col` is in the list of features.
This list also does not include the original name of features that were encoded.

For example, imagine `df` has columns "a" and "b", where "a" can be "1" or "2". We one-hot encode "a", so that columns "fklearn_feat__a==1" and "fklearn_feat__a==2" will be generated (and "a" will be removed).
Now, if one calls `xgb_classification_learner` with `features = ["a", "b"]` and `encode_extra_cols = True`, xgbc will actually consider as features the columns `["b", "fklearn_feat__a==1", "fklearn_feat__a==2"]`.

This might not be the best solution or the most elegant. Feedbacks are welcome.

### Todo list
- [x] Documentation
- [x] Tests added and passed


caique-lima:
@vultor33 any opinions regarding this approach? The idea is that with `encode_extra_cols` we are able to track from the original features which ones passed through a ohe(or any other learner that output columns in the format `fklearn_{feature_name}=={value}`)

PS: Feel free to review the code

caique-lima:
Relates to #68

vultor33:
> 
> 
> Relates to #68

No, no, this solution was really nice.

PR #65, opened 2019-05-13
Add extractor for temporal split evaluator

dieggoluis:
### Status
**READY**

### Todo list
- [x] Tests added and passed

### Background context
Although we have `temporal_split_evaluator` in fklearn, there is no extractor for this evaluator.

This PR adds a temporal split extractor for the corresponding evaluator.


src/fklearn/metrics/pd_extractors.py
```diff
@@ -24,7 +26,27 @@ def split_evaluator_extractor_iteration(split_value, result, split_col, base_ext
...
+    split_values = []
+    for key in split_keys:
+        date = key.split(eval_name)[1][1:]
+        try:
+            # just check time format
+            datetime.strptime(date, time_format)
+            split_values.append(date)
+        except ValueError:
+            # this might happen if result has temporal splitters using different data formats
+            pass
```
darlansf1: Shall we maybe raise a warn?

dieggoluis: I don't think so. Let's say you have 2 time split evaluators, one weekly and the other daily. It's a "normal" use case. This extractor would extract only the one which matches the given `timeformat`

PR #71, opened 2019-05-27
Changing the nan column to not include the first one dropped

jessicadesousa:
### Status
**READY**

### Background context
In the one hot encoder function, when hardcode_nans = True and drop_first_column = True too, the nans columns ended including the first column that was deleted. 

To solve this problem, I just changed the vector with the mapping to delete the first column after create the nan column.


dieggoluis:
Circleci is failing due to "ImportError: cannot import name 'factorial'". Shoud we fix scipy=1.2.1 on fklearn requirements? @caique-lima

PR #81, opened 2019-07-22
Add shap multiclass support

gbakie:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Fklearn supports returning SHAP values on prediction for binary classification and regression, but not for multiclass classification, even though it is supported by SHAP. This PR adds shap multiclass support, using the same approach for naming shap values columns as the one used for prediction columns.

### DIFF
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -39,7 +39,7 @@ def logistic_classification_learner(df: pd.DataFrame,
 
     target : str
         The name of the column in `df` that should be used as target for the model.
-        This column should be binary, since this is a classification model.
+        This column should be discrete, since this is a classification model.
 
     params : dict
         The LogisticRegression parameters in the format {"par_name": param}. See:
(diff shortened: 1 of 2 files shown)


src/fklearn/training/classification.py
```diff
@@ -188,16 +188,24 @@ def p(new_df: pd.DataFrame, apply_shap: bool = False) -> pd.DataFrame:
...
-            shap_values = list(explainer.shap_values(new_df[features]))
+            shap_values = explainer.shap_values(new_df[features])
             shap_expected_value = explainer.expected_value
 
-            shap_output = {"shap_values": shap_values,
-                           "shap_expected_value": np.repeat(shap_expected_value, len(shap_values))}
-
             if params["objective"] == "multi:softprob":
-                raise NotImplementedError("SHAP values are not implemented for multiclass XGBoost in fklearn")
+                shap_values_multiclass = {"shap_values_" + str(class_index): list(value)
```
prcastro: I would prefer the more pythonic form `f"shap_values_{class_index}"`

PR #93, opened 2019-09-02
Create multiple requirements to avoid install unused packages

caique-lima:
### Status
**IN DEVELOPMENT**

### Todo list
- [ ] Documentation

### Background context
As we add new models, this package gets bigger, what can lead to a bad user experience. So we want to find a way to enable user install only the dependencies he/she need

We introduce different requirements file based on scope, so if you want to use only `xgboost`, you can install only it using `pip install fklearn[xgboost]`, same for other models

- Are we ok with this change, or should we keep `all_deps`(except for test) as default?


caique-lima:
@marcelogdeandrade 

> * I think this creates a new step that makes it more painful and less scalable to add new models (create a new `requirements` file, add its reference to `setup.py` on two places)

Yes, but I don't see as "painful", neither as less scalable, as we don't use to add a lot of new dependencies, what we can do is use glob to find all files named as `requirements_*.txt` and based on that create the extra requires

> * We create another UX problem, which people will need to check if they have the required libraries for each method they are calling.

You don't need to know the requirements for each method, only for the model you're using, that's why I created the requirements groups(you can even just install `[all]` if you want)

> * I don't know if the UX impact with a "large" dependencies file is relevant enough to go through these other problems

The impact is really big, using slow connection is always a pain to install fklearn, users always have bigger packages than they need(libraries like xgboost, lgbm usually are big), sometimes you even need to install non python packages(lgbm with libomp).

> What do you think about this? I'm trying to find some dependency management references in other open-source projects as well.

Here some examples of similar projects doing this:

https://github.com/apache/airflow/blob/master/setup.py#L145
https://github.com/jupyter/notebook/blob/master/setup.py#L116
https://github.com/django/django/blob/master/setup.py#L87
https://github.com/uber/petastorm/blob/master/setup.py#L46

It's a common approach to modular libraries

marcelogdeandrade:
> @marcelogdeandrade
> 
> > * I think this creates a new step that makes it more painful and less scalable to add new models (create a new `requirements` file, add its reference to `setup.py` on two places)
> 
> Yes, but I don't see as "painful", neither as less scalable, as we don't use to add a lot of new dependencies, what we can do is use glob to find all files named as `requirements_*.txt` and based on that create the extra requires
> 
> > * We create another UX problem, which people will need to check if they have the required libraries for each method they are calling.
> 
> You don't need to know the requirements for each method, only for the model you're using, that's why I created the requirements groups(you can even just install `[all]` if you want)
> 
> > * I don't know if the UX impact with a "large" dependencies file is relevant enough to go through these other problems
> 
> The impact is really big, using slow connection is always a pain to install fklearn, users always have bigger packages than they need(libraries like xgboost, lgbm usually are big), sometimes you even need to install non python packages(lgbm with libomp).
> 
> > What do you think about this? I'm trying to find some dependency management references in other open-source projects as well.
> 
> Here some examples of similar projects doing this:
> 
> https://github.com/apache/airflow/blob/master/setup.py#L145
> https://github.com/jupyter/notebook/blob/master/setup.py#L116
> https://github.com/django/django/blob/master/setup.py#L87
> https://github.com/uber/petastorm/blob/master/setup.py#L46
> 
> It's a common approach to modular libraries

Makes sense, I didn't know that those libs could have such impact. I agree with your points about my comment.

pvcrossi:
As for the UX issue Marcelo described, would it be possible to make the default installation (without  any args) complete? 

This way, if the user is worried about dependency issues (or even connection/storage/whatever), they can declare whatever they want. But if they don't specify anything, they'll have everything

gcbeltramini:
The default installation comes from `install_requires=core_deps`, which I think is enough. If the user wants additional dependencies, they can install for example with
```shell
python -m pip install fklearn[xgboost,lgbm]
```

gcbeltramini:
Another repo that does this is airflow: https://github.com/apache/airflow/blob/a8dc50db4bc71be604ce44404fa6aedf0b81f484/setup.py#L145-L289

And I also prefer the idea of having multiple `requirements_...txt` files instead of having everything in the `setup.py` file.

gcbeltramini:
You need to edit https://github.com/nubank/fklearn/blob/master/MANIFEST.in:
```include requirements*.txt```

File: requirements_demos.txt [hunk 1]
```diff
@@ -0,0 +1,3 @@
+matplotlib>=3.0.2,<3.1.0
+tqdm>=4.32.1,<4.33.0
```
gabrielferreira95: This range of version for `tqdm` creates incompatibilities with `swifter`.

When doing `pip install fklearn` on a clean env, I get the following error:
`ERROR: swifter 0.293 has requirement tqdm>=4.33.0, but you'll have tqdm 4.32.2 which is incompatible.`

So probably it is a good idea to allow a higher range for `tqdm`.

As of today, the latest version of `tqdm` is [4.35.0](https://pypi.org/project/tqdm/). This solves the problem:
`tqdm>=4.32.1,<4.36.0`

PR #97, opened 2019-09-20
Fix/split evaluator extractor key error

### LINKED ISSUE #96: split_evaluator_extractor throws KeyError if not all values are present
### Problem description

Whenever an innermost extractor doesn't find an split_value that is not defined on a leaf (ie — it exists for some splits, but not all of them) it throws out a dict `KeyError`. 

### Code sample

```python
from fklearn.data.datasets import make_tutorial_data

from fklearn.metrics.pd_extractors import split_evaluator_extractor, evaluator_extractor
from fklearn.validation.evaluators import split_evaluator, r2_evaluator
import numpy as np

data = make_tutorial_data(50).dropna(subset=["feature3"]).assign(prediction=lambda d: d.target)

ev = split_evaluator(eval_fn=r2_evaluator, split_col="feature3")
ev2 = split_evaluator(eval_fn=ev, split_col="date")

results = ev2(data)

date_values = [
    np.datetime64("2015-01-06T00:00:00.000000000"),
    np.datetime64("2015-01-14T00:00:00.000000000"),
    np.datetime64("2015-01-22T00:00:00.000000000"),
    np.datetime64("2015-01-30T00:00:00.000000000"),
    np.datetime64("2015-03-08T00:00:00.000000000"),
    np.datetime64("2015-03-09T00:00:00.000000000"),
    np.datetime64("2015-04-04T00:00:00.000000000"),
]

e = evaluator_extractor(evaluator_name="r2_evaluator__target")
e1 = split_evaluator_extractor(base_extractor=e, split_col="feature3", split_values=["a", "b"])
e2 = split_evaluator_extractor(base_extractor=e1, split_col="date", split_values=date_values)

results_df = e2(results)  # this line throws a KeyError
```

### Expected behavior

Instead of throwing an error, it should return a pandas dataframe filled with `NaN`s for whenever a value is not defined in this particular split, but is defined on its parent.

### Possible solutions

`evaluator_extractor` should return `nan` when the value is not found.

gppeixoto:
### Status
**READY**

### Todo list
- [x] Issue: closes #96 

`split_evaluator_extractor` will return `nan` for when it doesn't find a metric for a split value that passed on arguments instead of crashing with KeyError.


gabrielvcbessa:
Think it is interesting to add a test for that

gppeixoto:
> Think it is interesting to add a test for that

you're correct, thanks. added at 2041012

PR #98, opened 2019-09-23
Fix/split evaluator extractor key error

gppeixoto:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes #97 

### Background context
When `split_values` is passed on `split_evaluator_extractor` and it doesn't find a key within the results dictionary tree, it throws a key error, even if the other splits are defined.

Returns `nan` when any of the split values passed on argument is not present.

### Changes

* f42558e fix(pd_extractors.py): return nan if split value is not found
* 9f3a192 chore(VERSION): bump patch
* c954024 docs(CHANGELOG.md): update changelog
* c68afb8 test(test_pd_extractors.py): test when split value is missing for split_evaluator


tests/metrics/test_pd_extractors.py
```diff
@@ -4,20 +4,94 @@
...
                                            spearman_evaluator, split_evaluator,
-                                           temporal_split_evaluator)
+                                           temporal_split_evaluator, mse_evaluator)
 from fklearn.validation.splitters import (
     forward_stability_curve_time_splitter, out_of_time_and_space_splitter,
     stability_curve_time_splitter, time_learning_curve_splitter)
 from fklearn.validation.validator import validator
 
 
+def test__split_evaluator_extractor__when_split_value_is_missing():
```
faciolihenrique: This test doesn't cover the case of a three depth tree, does it? Either way, using the `{}` guarantee any tree will work

gppeixoto: works for N

caique-lima:
As this is a bug fix to `1.15.x` change the target branch to `1.15.x` instead of master branch

caique-lima:
@gppeixoto can you remove the changes in the changelog + version? Seems that you created this based on the master branch, and not the `1.15.x`, this way this specific PR will be the fix for the version `1.16.0` and not the version `1.15.x`

PR #101, opened 2019-10-01
Adding PR AUC evaluator

dieggoluis:
### Status
**READY**

### Todo list
- [ ] Documentation
- [ ] Tests added and passed

### Background context
Even though we have ROC AUC evaluator implemented in `fklearn`, we are still missing PR AUC evaluator.

This PR:
1) Adds a new evaluator `pr_auc_evaluator` 
2) Changes the generic auc evaluator name from `auc_evaluator` to `roc_auc_evaluator`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,9 @@ venv.bak/
 .spyderproject
 .spyproject
 
+# Visual Studio Code
+.vscode
+
 # Rope project settings
 .ropeproject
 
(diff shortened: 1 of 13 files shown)


caique-lima:
@dieggoluis what do you think about create a method `auc_evaluator` that calls the `roc_auc_evaluator` and raises a warning telling the users that we will drop this method in the next major release(`2.0.0`)? Personally I think it's a good way to do a soft rollout of the change, otherwise we will need to release the 2.0.0 to support this api change

dieggoluis:
@caique-lima it's actually a great idea! How about previous methods/tests etc that uses `auc_evaluator` and I changed them to `roc_auc_evaluator`. Should I keep them calling `auc_evaluator`, given that I'll create this method `auc_evaluator` compatible with the current version?

caique-lima:
@dieggoluis for me it's ok to keep the changes you did

PR #107, opened 2019-10-31
Fix space_time_split_dataset

pvcrossi:
### Status
**READY**

### Todo list
- [x] Tests added and passed

### Background context
Function `preprocessing.splitting.space_time_split_dataset` is returning faulty holdout datasets.

Since all datasets are defined in terms of `holdout_space` (which is a random subset of "ids" found in the training period), all new "ids" appearing only in the outtime period will be wrongfully attributed to inspace.

In the limit, when there's just one data point per id, there'll be an inversion of `outtime_inspace` and `outtime_outspace`.

This PR fixes this by defining everything in terms of `inspace`.

### DIFF
--- a/src/fklearn/preprocessing/splitting.py
+++ b/src/fklearn/preprocessing/splitting.py
@@ -155,12 +155,13 @@ def space_time_split_dataset(dataset: pd.DataFrame,
         train_period_space = np.sort(all_space_in_time)
 
         # randomly sample accounts from the train period to hold out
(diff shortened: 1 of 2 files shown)


sadikneipp:
Nice catch. I feel this is the result of bad tests and we should invest some brain-time on building more test-cases (like the one described in this PR)

pvcrossi:
@sadikneipp I now added some tests for the two cases I could think of

PR #111, opened 2019-11-29
Fix shap

### LINKED ISSUE #109: Not working with newer SHAP versions
### Problem description
When running lgbm with `apply_shap=True`, it throws a `ValueError`. Apparently there was a change on Shap's output (to a list of ndarray) on its newer versions. (https://github.com/slundberg/shap/blob/master/shap/explainers/tree.py#L194)

### Issue location
fklearn/training/classification.py on lgbm_classification_learner

### Expected behavior
`f(df, apply_shap=True)` used to return the original dataframe with a column "shap_values" with an array as value

### Possible solutions
Adapt lgbm_classication_learner to newer (>=0.31.0) shap versions

tatasz:
### Status
**On Hold**

### Todo list
- [x] Issue: closes #109

### Background context
There was a change on Shap's output (to a list of ndarray) on its newer versions, and the classification learners throw a `ValueError` when `apply_shap=True` for binary classification. Multiclass classification and regression are not affected.

Uses `shap_values[1]` and `explainer.expected_value[1]` for binary classification.


sadikneipp:
Maybe fix/add some tests to ensure the new behaviour is working as expected? There are some tests related to shap failing on circleCI

tatasz:
> Maybe fix/add some tests to ensure the new behaviour is working as expected? There are some tests related to shap failing on circleCI

The thing is, im sort of dummy with fklearn tests and I am not sure why exactly they are failing (I'm a dummy and those tests are kinda complex). Will try to look into it again, but cannot guarantee I can fix it :(

robotenique:
Try using checking before assuming it is a list, something like

```python
    if isinstance(expected_value, list):
        expected_value = expected_value[1]
    if isinstance(shap_values,  list):
        shap_values = shap_values[1]
   return expected_value, shap_values
```

tatasz:
> Try checking the type before assuming it is a list, something like
> 
> ```python
>     if isinstance(expected_value, list):
>         expected_value = expected_value[1]
>     if isinstance(shap_values,  list):
>         shap_values = shap_values[1]
>    return expected_value, shap_values
> ```
> 
> I think it may help with the tests

Not really. We kind of know it is a list, so I would need to understand when it is not a list before doing this.

robotenique:
> > Try checking the type before assuming it is a list, something like
> > ```python
> >     if isinstance(expected_value, list):
> >         expected_value = expected_value[1]
> >     if isinstance(shap_values,  list):
> >         shap_values = shap_values[1]
> >    return expected_value, shap_values
> > ```
> > 
> > 
> > I think it may help with the tests
> 
> Not really. We kind of know it is a list, so I would need to understand when it is not a list before doing this.

We don't know. In fact, you didn't change the shap version in the requirements, which is still `shap>=0.28.5,<0.29.0`. This means that the shap_values still returns a single value. By adding this check you maintain backwards compatibility while also working as expected in newer SHAP versions.

Or you could change the shap version itself in `requirements_tools.txt`

tatasz:
I actually changed the shap version.

robotenique:
> I actually changed the shap version.

That's weird, the circleCI tests are failing with `IndexError: invalid index to scalar variable.` which occur when trying to index something that isn't a list...

tatasz:
> > I actually changed the shap version.
> 
> That's weird, the circleCI tests are failing with `IndexError: invalid index to scalar variable.` which occur when trying to index something that isn't a list...

Yep this is what im trying to understand. It got the new shap version, so it should be a list, but for some reason is not, which means something is pretty broken there :(

tatasz:
Ok, there is an issue with Shap vs catboost, which is still open: https://github.com/slundberg/shap/issues/753
https://github.com/slundberg/shap/issues/750

luamct:
@tatasz Can we close this PR until we have a strategy to deal with this?

tatasz:
> @tatasz Can we close this PR until we have a strategy to deal with this?

Yep. I considered using shap from catboost, but had no time to work on that so...

PR #113, opened 2020-01-02
Calibration parameters

ValeriaGomes:
### Status
**READY**

### Todo list
- [ ] Documentation
- [ ] Tests added and passed

### Background context
Limits of 0 and 1 restrict fklearn's isotonic calibration to be used only on classification problems, but calibration may be important for regression models too. 

Adding upper and lower bounds increases the calibration flexibility, but the default values are the same for classification problems, so it would not break the codes build with the former version.

'src/fklearn/training/calibration.py'


src/fklearn/training/calibration.py
```diff
@@ -13,31 +13,31 @@
 def isotonic_calibration_learner(df: pd.DataFrame,
                                  target_column: str = "target",
                                  prediction_column: str = "prediction",
-                                 output_column: str = "calibrated_prediction") -> LearnerReturnType:
+                                 output_column: str = "calibrated_prediction",
+                                 y_min: float = 0,
+                                 y_max: float = 1) -> LearnáerReturnType:'
```
arthurpaulino: Typo on `LearnáerReturnType` (á)

src/fklearn/training/calibration.py
```diff
@@ -57,4 +57,4 @@ def p(new_df: pd.DataFrame) -> pd.DataFrame:
     return p, p(df), log
 
 
-isotonic_calibration_learner.__doc__ += learner_return_docstring("Isotonic Calibration")
+isotonic_calibration_learner.__doc__ += learner_return_docstring("Isotonic Calibration")
```
arthurpaulino: This is just polishing, but can you add the newline?

src/fklearn/training/calibration.py
```diff
@@ -13,10 +13,12 @@
...
                                  target_column: str = "target",
                                  prediction_column: str = "prediction",
-                                 output_column: str = "calibrated_prediction") -> LearnerReturnType:
+                                 output_column: str = "calibrated_prediction",
+                                 y_min: float = 0,
+                                 y_max: float = 1) -> LearnerReturnType:
     """
     Fits a single feature isotonic regression to the dataset.
-
+
```
arthurpaulino: These white spaces are giving lint error as well

ValeriaGomes: Thanks!

src/fklearn/training/calibration.py
```diff
@@ -13,7 +13,9 @@
 def isotonic_calibration_learner(df: pd.DataFrame,
                                  target_column: str = "target",
                                  prediction_column: str = "prediction",
-                                 output_column: str = "calibrated_prediction") -> LearnerReturnType:
+                                 output_column: str = "calibrated_prediction",
+                                 y_min: float = 0,
+                                 y_max: float = 1) -> LearnerReturnType:
```
caique-lima: Maybe it's good to set `0.0` and `1.0`

ValeriaGomes: 🆗

caique-lima:
I'm not sure why it didn't failed before, but you can fix changing this row

https://github.com/nubank/fklearn/blob/master/src/fklearn/validation/validator.py#L152

to

```
return args['cols'] if args else []
```

Basically the error is telling us that `args` may be `None`, and if this happens our code would fail, with this fix we would always return something

PR #119, opened 2020-01-14
Issue 118 catboost

### LINKED ISSUE #118: Allow fit params for CatBoost
- CatBoost has several parameters inside the `fit` method. One possible solution is to add a `fit_params` into the CatBoost function in FkLearn for increased flexibility.
- This improvement would have to be done for both classification and regression.

This is important especially for CatBoost because the parameter that allows it to treat categorical features byitself is inside the `fit` method, and this is the main differential in this library.

We would be able to configure the fit function using FkLearn.

### Additional Information
https://catboost.ai/docs/concepts/python-reference_catboost_fit.html

### LINKED ISSUE #188: Code style preferences
Hey everyone!

I'm looking at the code and I saw that there are some inconsistencies between code formatting, mainly in the usage of quotes (`"` on some strings and `'` in other cases).
So what do you think about establishing some preferences on code style? I mean, we already have flake8 to check whether the code follows some PEP8 recommendations, but I want to be sure before adding extra elements or something that could break the current approach?

I know that we have [this section](https://fklearn.readthedocs.io/en/latest/contributing.html#code-standards) and that we are following [this guide](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#313-imports-formatting) and PEP8, so I just want to be sure that we are in the same channel.

Also if we don't want to start adding a lot of manual/custom rules, we can work directly with an unopinionated framework like [black](https://github.com/psf/black), which usually follows everything in the PEP8 (except for the double quotes).

Thanks in advance!

ValeriaGomes:
### Status
**READY**

### Todo list
- [ ] Documentation
- [ ] Tests added and passed
- [ ] Issue: closes #118 

### Background context
- CatBoost has several parameters inside the `fit` method. One possible solution is to add a `fit_params` into the CatBoost function in FkLearn for increased flexibility.
- This improvement would have to be done for both classification and regression.

This is important especially for CatBoost because the parameter that allows it to treat categorical features by itself is inside the `fit` method, and this is the main differential in this library.

Allow the user to set fit params in CatBoost: https://catboost.ai/docs/concepts/python-reference_catboostregressor_fit.html#python-reference_catboostregressor_fit

src/fklearn/training/classification.py
src/fklearn/training/regression.py


src/fklearn/training/classification.py
```diff
@@ -290,7 +291,14 @@ def catboost_classification_learner(df: pd.DataFrame,
 
     encode_extra_cols : bool (default: True)
         If True, treats all columns in `df` with name pattern fklearn_feat__col==val` as feature columns.
+
+    fit_params: dict, optional
+        Dictionary in the format {"hyperparameter_name" : hyperparameter_value.
```
gabrielferreira95: I think this line is missing a closing `}`

src/fklearn/training/regression.py
```diff
@@ -256,6 +257,12 @@ def catboost_regressor_learner(df: pd.DataFrame,
 
     weight_column : str, optional
         The name of the column with scores to weight the data.
+
+    fit_params: dict, optional
+        Dictionary in the format {"hyperparameter_name" : hyperparameter_value.
```
gabrielferreira95: Same here. Missing a `}`.

ValeriaGomes:
I'm reviewing [this test](https://github.com/nubank/fklearn/blob/issue-118-catboost/tests/training/test_regression.py#L200) because it's breaking when I insert the `fit_params`.

caique-lima:
Closing this one as we don't have updates

PR #120, opened 2020-01-14
Fix shap

### LINKED ISSUE #109: Not working with newer SHAP versions

tatasz:
### Status
**READY**

### Todo list
- [x] Tests added and passed
- [x] Issue: closes #109 

### Background context
Update shap to be compatible with latest shap versions

Updated lgbm shap, and changed the way multiclass shap is calculated for catboost.
UPD: added description for magic numbers

src/fklearn/training/classification.py


dieggoluis:
Is there any test covering those changes? 🤔

tatasz:
> Is there any test covering those changes? 🤔

I may be wrong, but i think they are covered by the already existing test, since the changes don't add anything new, but only adapt the shap output from newer shap versions to our format.

dieggoluis:
Ok, make sense, so tests would fail if you just update shap version without adapting the code. Got it.

PR #121, opened 2020-01-14
Columns duplicator learner and decorator

### LINKED ISSUE #116: Output encoded values into a new column
Currently, most of fklearn transformations (such as `transformation.target_categorizer`, `transformation.truncate_categorical`, `transformation.rank_categorical`) replace the original values by the encoded values.

However, for a series of reasons (such as preserving the original value, or applying multiple encodings to the same feature), one may want to save the encoded values as a different feature, and keep the original column.

It would be useful to add an option to save the encoded feature to a new column.

vitorsrg:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [X] Issue: closes #116 

- Add a new transformer `feature_duplicator`
- Add new meta decorator `column_duplicatable` to transformers and learners to duplicate columns

### Background
Currently, the transformations (such as `target_categorizer`) replace the value in the column. A better approach would be to allow the user to preserve the original value, outputting the encoded feature to a new column. This would also enable users to apply more than one transformation to the same column (example: frequency and target encoding of the same feature) without needing to duplicate the column first.

Usage examples in comments.


vitorsrg:
sample test

```
df2 = pd.DataFrame(np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]), columns=['a', 'b', 'c'])

print(feature_duplicator(df2.copy(), columns_to_duplicate=['b', 'c'], columns_mapping={'b': 'b2', 'c': 'c3'})[1])
print(feature_duplicator(df2.copy(), columns_to_duplicate=['b', 'c'], preffix='prefix______')[1])
print(feature_duplicator(df2.copy(), columns_to_duplicate=['b', 'a'], suffix='____suffix')[1])
print(feature_duplicator(df2.copy(), columns_to_duplicate=['c'], preffix='before___', suffix='___after')[1])

pipe = target_categorizer(
columns_to_categorize=['b', 'a'], target_column='c', columns_to_duplicate=['b', 'a'], suffix='____suffix')
p, p_df, log = pipe(df2.copy())

print(p_df)
print(df2)
print(p(df2.copy()))
```

vitorsrg:
sample output

```
   a  b  c  b2  c3
0  1  2  3   2   3
1  4  5  6   5   6
2  7  8  9   8   9
   a  b  c  prefix______b  prefix______c
0  1  2  3              2              3
1  4  5  6              5              6
2  7  8  9              8              9
   a  b  c  b____suffix  a____suffix
0  1  2  3            2            1
1  4  5  6            5            4
2  7  8  9            8            7
   a  b  c  before___c___after
0  1  2  3                   3
1  4  5  6                   6
2  7  8  9                   9
   a  b  c  b____suffix  a____suffix
0  1  2  3          4.5          4.5
1  4  5  6          6.0          6.0
2  7  8  9          7.5          7.5
   a  b  c
0  1  2  3
1  4  5  6
2  7  8  9
   a  b  c  b____suffix  a____suffix
0  1  2  3          4.5          4.5
1  4  5  6          6.0          6.0
2  7  8  9          7.5          7.5
```

vitorsrg:
sample test:

```python
df2 = pd.DataFrame(np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]), columns=['a', 'b', 'c'])

pipe = target_categorizer(
columns_to_categorize=['b', 'a'], target_column='c', columns_to_duplicate=['b', 'a'], suffix='____suffix')

print(df2)
print(pipe(df2.copy())[1])
```

expected output:

```
   a  b  c
0  1  2  3
1  4  5  6
2  7  8  9
   a  b  c  b____suffix  a____suffix
0  1  2  3          4.5          4.5
1  4  5  6          6.0          6.0
2  7  8  9          7.5          7.5
```

vitorsrg:
TODO: refactor use cases

```
target_categorizer(columns_to_categorize=['b', 'a'], target_column='c') # inplace!
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], suffix='____suffix')
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], preffix='preffix___')
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], columns_mapping={...})
```

vitorsrg:
use cases

```
target_categorizer(target_column='c', columns_to_categorize=['b', 'a']) # inplace!
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], suffix='____suffix')
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], preffix='preffix___')
target_categorizer(target_column='c', columns_to_categorize=['b', 'a'], columns_mapping={...})
```

src/fklearn/preprocessing/schema.py
```diff
@@ -0,0 +1,121 @@
...
+    suffix: int (default None)
+        Suffix to add to columns to duplicate
+
+    Returns
+    ----------
+    increased_dataset : pandas.DataFrame
+        A dataset with repeated columns
+    """
+
+    stackname = inspect.stack()[0].function
```
prcastro: It isn't very clear for me what this do just reading the code. What about leaving at least a comment or a more descriptive variable name? Also, the use of `inspect` and `eval` is pretty unsafe, isn't there another way of achieving the same thing here?

src/fklearn/training/transformation.py
```diff
@@ -51,6 +53,7 @@ def p(new_data_set: pd.DataFrame) -> pd.DataFrame:
 selector.__doc__ += learner_return_docstring("Selector")
 
 
+@column_duplicatable('columns_to_cap')
```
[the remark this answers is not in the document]
vitorsrg: Sorry, i couldn't understand what do you mean

vitorsrg:
> Hey! Nice job! I left some tips in some parts of the code. 

😄 

> Also, I suggest you write a Background section on the PR, explaining what are the use case of this new features (maybe even with an example)

What is a Background section?
The use cases and examples are given above in comments to this PR

prcastro:
> > Hey! Nice job! I left some tips in some parts of the code.
> 
> 😄
> 
> > Also, I suggest you write a Background section on the PR, explaining what are the use case of this new features (maybe even with an example)
> 
> What is a Background section?
> The use cases and examples are given above in comments to this PR

I meant the reasoning behind this PR. What problem is it trying to solve? Can you exemplify the problem?

src/fklearn/training/transformation.py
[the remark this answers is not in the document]
prcastro: If you add the decorator to the current tests, we are loosing the test for the case where there is no decorator. We should test each of those transformation with AND without the decorator.

tatasz:
> > > Hey! Nice job! I left some tips in some parts of the code.
> > 
> > 
> > 😄
> > > Also, I suggest you write a Background section on the PR, explaining what are the use case of this new features (maybe even with an example)
> > 
> > 
> > What is a Background section?
> > The use cases and examples are given above in comments to this PR
> 
> I meant the reasoning behind this PR. What problem is it trying to solve? Can you exemplify the problem?

Is this needed? I mean, it is basically closing 

> > > Hey! Nice job! I left some tips in some parts of the code.
> > 
> > 
> > 😄
> > > Also, I suggest you write a Background section on the PR, explaining what are the use case of this new features (maybe even with an example)
> > 
> > 
> > What is a Background section?
> > The use cases and examples are given above in comments to this PR
> 
> I meant the reasoning behind this PR. What problem is it trying to solve? Can you exemplify the problem?

Added

src/fklearn/preprocessing/schema.py
```diff
@@ -0,0 +1,114 @@
...
+    return p, p(df), log
+
+
+def column_duplicatable(columns_to_bind: str) -> Callable:
+    """
+    Decorator to duplicate some columns in the dataframe
+
+    Parameters
+    ----------
+    columns_to_bind: list of str
```
prcastro: Is it a `list of str` or a `str`

vitorsrg: It's wrong, it's just for str

src/fklearn/preprocessing/schema.py
```diff
@@ -0,0 +1,115 @@
...
+import toolz
+
+from typing import Any, Callable, Dict, List, Optional, Union
+from fklearn.types import LearnerLogType, LearnerReturnType
+
+
+@toolz.curry
+def feature_duplicator(df: pd.DataFrame,
+                       columns_to_duplicate: Optional[List[str]] = None,
+                       columns_mapping: Optional[Dict[str, str]] = None,
```
kaiubiferreira: What's the idea for this parameter? 
Is it a way for the user to have a more flexible mapping than prefix/suffix?
I couldn't find where this is used (or tested)

vitorsrg: yes, that's the idea
hm... i'll add them

src/fklearn/preprocessing/schema.py
```diff
@@ -0,0 +1,134 @@
...
+@toolz.curry
+def feature_duplicator(
+    df: pd.DataFrame,
+    columns_to_duplicate: Optional[List[str]] = None,
+    columns_mapping: Optional[Dict[str, str]] = None,
+    preffix: Optional[str] = None,
+    suffix: Optional[str] = None
+) -> LearnerReturnType:
+    """
+    Duplicates some columns in the dataframe
```
caique-lima: Would be cool to provide are clearer docstring, explaining what is happening and giving examples(same for the `column_duplicatable` method)

src/fklearn/training/transformation.py
```diff
@@ -512,6 +517,7 @@ def p(new_df: pd.DataFrame) -> pd.DataFrame:
 rank_categorical.__doc__ += learner_return_docstring("Rank Categorical")
 
 
+@column_duplicatable('columns_to_categorize')
```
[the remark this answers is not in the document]
vitorsrg: The idea for the decorator was that multiple learners are changing values inplace, so ideally we should duplicate those columns before changing them inplace

The problem is that copying and pasting this behaviour is very cumbersome and error-prone, so the decorator just appends the duplicator before the problematic learner

Also the decorator is easier for users, so they dont have to handle columns lists when defining the pipeline

All the existing tests worked with the default behaviour, so it seams to be safe for most use cases

src/fklearn/preprocessing/schema.py
vitorsrg: Is there an example on which i can ground?

src/fklearn/training/transformation.py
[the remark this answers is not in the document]
caique-lima: 🤔 ok!

After some tests I think now it's clear what is the impact of the change. The decorator isn't doing anything to the learners, except saying that they are able to duplicate columns.

A proper Docstring on the decorators definition will definitely help. Now that everything is clear, let's see what we can do to merge this, I have some recommendations:

- Docstrings(I will respond the other comment with some examples)
- Documentation, maybe add a example of how to use this feature in the documentation can help(not required but could help the users, maybe something like [this](https://fklearn.readthedocs.io/en/latest/examples/learning_curves.html))
- Test for the affected methods, without this is hard to guarantee that the behaviour will always be the same, together with the regular test of the learners, I would add a new one that test the behaviour using the decorator for each learner we applied it(with a prefix or suffix)

src/fklearn/preprocessing/schema.py
caique-lima: I don't have a decent example, as we don't have a lot of other stuff doing similar things, but it could be more complete like the following:

https://github.com/nubank/fklearn/blob/master/src/fklearn/validation/splitters.py#L264
https://github.com/nubank/fklearn/blob/master/src/fklearn/validation/splitters.py#L389
https://github.com/nubank/fklearn/blob/master/src/fklearn/training/pipeline.py#L13

In your case would be cool to have an example explaining exactly the behaviour, this way people don't need to see this PR to understand what is happening

src/fklearn/preprocessing/schema.py
```diff
@@ -0,0 +1,134 @@
...
+        'feature_duplicator': {
+            'columns_to_duplicate': columns_to_duplicate,
+            'columns_mapping': columns_mapping,
+            'preffix': preffix,
+            'suffix': suffix,
+            'columns_final_mapping': columns_final_mapping,
+        }
+    }
+
+    return p, p(df), log
```
caique-lima: maybe here we should copy the Data Frame `df` otherwise this method will mutate the original dataset

caique-lima:
Closing this one due to no recent updates

src/fklearn/training/transformation.py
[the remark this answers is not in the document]
tatasz: Done

tatasz:
UPD:

tests/training/test_transformation.py
```diff
@@ -5,27 +5,36 @@
 from numpy import nan, round, sqrt, floor, log as ln
```
caique-lima, line 5: 🙈  it is hard to track what happened with this file, from what I understood:
- Some code style
- Add new tests to existing ones covering the use cases with prefix/suffix on transformers

caique-lima (approved):

prcastro (approved):

PR #122, opened 2020-01-15
Pipeline logs support for multiple learners with the same name

dieggoluis:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
The logs of the pipeline `build_pipeline` do not support multiples learners with the same name. If there are two learners with the same name (E.g: multiple xgboost learners each one using the same features, but different targets. In those cases) the pipeline logs would store only information regarding the last repeated learner.

This PR creates a new pipeline `build_pipeline_repeated_learners` that will support multiples learners with the same name.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.17.0]
+- **New**
+    - Add a new pipeline build_pipeline_repetead_learners that supports multiple learners with the same name.
+
```
caique-lima: You can remove this one as we update the changelog in the release PR with all the changes.

tests/training/test_pipeline.py
```diff
@@ -6,12 +6,13 @@
...
 
 from fklearn.training.imputation import placeholder_imputer
-from fklearn.training.pipeline import build_pipeline
+from fklearn.training.pipeline import build_pipeline, build_pipeline_repeated_learners
 from fklearn.training.regression import xgb_regression_learner
 from fklearn.training.transformation import count_categorizer, onehot_categorizer
 
 
-def test_build_pipeline():
+@pytest.mark.parametrize("pipeline_fn", [build_pipeline, build_pipeline_repeated_learners])
```
caique-lima: Nice approach with the tests 👏

src/fklearn/training/pipeline.py
```diff
@@ -7,10 +8,33 @@
...
+                         "Make sure no predict function uses arguments like *args or **kwargs."
+                         .format(learner.__name__, ', '.join(var_args)))
+
+
+def _check_unfilled_arg(*learners: LearnerFnType) -> None:
+    for l in learners:
+        _has_one_unfilled_arg(l)
+
+
+def build_pipeline_repeated_learners(*learners: LearnerFnType) -> LearnerFnType:
```
caique-lima: This doesn't look like the best approach, as we may need to change the same code in two places. From what I've saw we only change two lines, the one with the `defaultdict`, and the one where we append the logs. I think we can rewrite this in a better way to avoid code duplication.

I think you can add a parameter to the `build_pipeline` method that defaults to False, something like `has_repeated_learners`, and them use this to change the way we initialise the serialisation dict, and the way we append data. Could be something like:

```python
def build_pipeline(*learners: LearnerFnType, has_duplicated_learners: bool = False) -> LearnerFnType:
(....)
serialisation = defaultdict(list) if has_duplicated_learners else {}  # type: dict
(....)
if has_duplicated_learners:
    serialisation[learner_name].append({"fn": learner_fn, "log": learner_log, **model_objects})
else
    serialisation[learner_name] = {"fn": learner_fn, "log": learner_log, **model_objects}
```

This approach isn't a breaking change, and we avoid a lot of code duplication

WDYT?

CHANGELOG.md
dieggoluis: 👍

src/fklearn/training/pipeline.py
dieggoluis: Great idea! I just refactored the code and asked for your review again.

Issue #124, opened 2020-02-10
Missing features removal with SimpleImputer


vitorsrg:
### Code sample
In the sample code below, a column is removed from the dataset during the pipeline

```python
>>> from sklearn.impute import SimpleImputer
>>> import numpy as np
>>> imp = SimpleImputer()
>>> imp.fit([[0, np.nan], [1, np.nan]])
>>> imp.transform([[0, np.nan], [1, 1]])
array([[0.],
       [1.]])
```

### Problem description
Currently `sklearn.impute.SimpleImputer` silently removes features that are `np.nan` on every training sample.

Therefore https://github.com/nubank/fklearn/blob/06475b6c425e0098e60b367d30de7272dbe96727/src/fklearn/training/imputation.py#L43 fails as `new_data.shape[1] != len(columns_to_impute)`. 

### Possible solutions
For the problematic features, either keep their values if valid or impute a default value during `transform`.

PR #126, opened 2020-02-12
Imputer fill_value

vitorsrg:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes https://github.com/nubank/fklearn/issues/124

- add `fill_value` to `imputer`
- add to `.gitignore` some files created by the tests


vitorsrg:
For reference: https://github.com/nubank/fklearn/pull/125

tests/training/test_imputation.py
```diff
@@ -6,25 +6,29 @@
 def test_imputer():
     input_df = pd.DataFrame({
         'col1': [10, 13, 10],
-        'col2': [50, 100, None]
+        'col2': [50, 100, None],
```
gabrielvcbessa: Instead of changing the test `test_imputer` I would suggest creating a new one, something like `test_imputer_with_fill_value`. 

This way we would have tests testing both scenarios.

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
...
+    fill_value : Any, (default=None)
+        if not None, use this as default value when some feature only contains NA values.
     """
 
+    columns_to_fill = list()
+    columns_imputable = columns_to_impute
+    if fill_value is not None:
+        df_is_nan = df[columns_to_impute].isna().all(axis=0)
+        columns_to_fill = list(df_is_nan[df_is_nan].index)
+        columns_imputable = list(filter(lambda column: column not in columns_to_fill, columns_to_impute))
```
gppeixoto: please , avoid declaring a variable and mutating state. one alternative is  to use python's ternary operator like `a = 1 if False else 2`. one way to achieve this is extracting this in another method (python downsides on being real bad to make private methods; so maybe it can be an "auxiliary" method inside this one)

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
...
 
     def p(new_data_set: pd.DataFrame) -> pd.DataFrame:
-        new_data = imp.transform(new_data_set[columns_to_impute])
-        new_cols = pd.DataFrame(data=new_data, columns=columns_to_impute).to_dict('list')
-        return new_data_set.assign(**new_cols)
+        new_df = new_data_set[columns_to_impute].copy()
+        new_df.loc[:, columns_imputable] = imp.transform(new_df[columns_imputable])
+        if columns_to_fill:
+            new_df.loc[:, columns_to_fill] = new_df.loc[:, columns_to_fill].fillna(value=fill_value)
+        return new_df
```
gppeixoto: same here, avoid mutating state. you can use the `assign` method to work around this; without needing to manually copy the dataset again using `.copy`

src/fklearn/training/imputation.py
prcastro: The following is cleaner and faster:

```python
[c for c in columns_to_impute if c not in columns_to_fill]
```

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
         - If "mean", then replace missing values using the mean along the axis.
         - If "median", then replace missing values using the median along the axis.
         - If "most_frequent", then replace missing using the most frequent value along the axis.
+
+    fill_value : Any, (default=None)
+        if not None, use this as default value when some feature only contains NA values.
```
gppeixoto: im not sure this is clear; is the expected behavior for `fill_value` to be filled *strictly only when 100%* of the feature values is `na`? what if there's only 1 not missing (or any number > 0, fwiw)? when should i use `fill_value`? i think the description can be improved here.

prcastro: If this is the default value, I would change this name to something like `default_fill_value` or something like this

prcastro: Also, I would use the `column` instead of `feature` and `missing` instead of `NA`:

```
if not None, use this as default value when one of the columns_to_impute only contains missing values.
```

src/fklearn/training/imputation.py
prcastro: Also,  I would name this variable something like `columns_to_use_strategy`.

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
...
+    imp.fit(df[columns_imputable].values)
 
     def p(new_data_set: pd.DataFrame) -> pd.DataFrame:
-        new_data = imp.transform(new_data_set[columns_to_impute])
-        new_cols = pd.DataFrame(data=new_data, columns=columns_to_impute).to_dict('list')
-        return new_data_set.assign(**new_cols)
+        new_df = new_data_set[columns_to_impute].copy()
+        new_df.loc[:, columns_imputable] = imp.transform(new_df[columns_imputable])
+        if columns_to_fill:
+            new_df.loc[:, columns_to_fill] = new_df.loc[:, columns_to_fill].fillna(value=fill_value)
```
prcastro: What happens here if the `columns_to_fill` have different types? I would prefer to have multiple `filll_values` (it could be a `Dict[column, value]`)

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
...
         - If "most_frequent", then replace missing using the most frequent value along the axis.
+
+    fill_value : Any, (default=None)
+        if not None, use this as default value when some feature only contains NA values.
     """
 
+    columns_to_fill = list()
+    columns_imputable = columns_to_impute
+    if fill_value is not None:
+        df_is_nan = df[columns_to_impute].isna().all(axis=0)
```
prcastro: I would name this `columns_all_missing` or `mask_all_missing`, since this isn't a dataframe (and therefore we shouldn't call it a `df_...`)

src/fklearn/training/imputation.py
```diff
@@ -32,23 +33,41 @@ def imputer(df: pd.DataFrame,
         - If "mean", then replace missing values using the mean along the axis.
         - If "median", then replace missing values using the median along the axis.
         - If "most_frequent", then replace missing using the most frequent value along the axis.
+
+    fill_value : Any, (default=None)
+        if not None, use this as default value when some feature only contains NA values.
     """
 
+    columns_to_fill = list()
+    columns_imputable = columns_to_impute
```
prcastro: I would find it easier to read if we define this in an `else` statement:

```
if fill_value:
   ...
else:
   columns_to_fill = list()
   columns_imputable = columns_to_impute
```

src/fklearn/training/imputation.py
vitorsrg: I think that keeping the default value outside of the `if` block is clearer, but I'll refactor to an `else` block

vitorsrg: which variable?

src/fklearn/training/imputation.py
vitorsrg: I'm trying (and failing) to mimic the `fill_value` description in https://scikit-learn.org/stable/modules/generated/sklearn.impute.SimpleImputer.html

src/fklearn/training/imputation.py
vitorsrg: done

src/fklearn/training/imputation.py
vitorsrg: `imputer`, `placeholder_imputer` and `SimpleImputer` don't accept multiple strategies, so for simplicity i think this feature also shouldn't

src/fklearn/training/imputation.py
vitorsrg: reverted to original code

tests/training/test_imputation.py
```diff
@@ -30,6 +30,32 @@ def test_imputer():
...
+    expected = pd.DataFrame({
+        'col1': [10.0, 13.0, 10.0],
+        'col2': [50.0, 100.0, 75.0],
+        'col3': [10.0, -999, -999]
+    })
+
+    pred_fn, data, log = imputer(input_df, ["col1", "col2", "col3"], "mean", placeholder_value=-999)
+
+    print(pred_fn(df))
+    print(expected)
```
gabrielvcbessa: Need to remove all prints.

vitorsrg: 🤦‍♂

tests/training/test_imputation.py
```diff
@@ -30,6 +30,30 @@ def test_imputer():
...
+
+    expected = pd.DataFrame({
+        'col1': [10.0, 13.0, 10.0],
+        'col2': [50.0, 100.0, 75.0],
+        'col3': [10.0, -999, -999]
+    })
+
+    pred_fn, data, log = imputer(input_df, ["col1", "col2", "col3"], "mean", placeholder_value=-999)
+
+    assert expected.equals(pred_fn(df))
```
gppeixoto: instead of using `assert` and `equals`; it's best to use directly the [assertions provided by pandas](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html). i'd change here to

```python
actual = pred_fn(df)
pd.testing.assert_frame_equal(expected, actual)
```

it also provides a more meaningful description whenever the assertion fails; instead of plain "assertion error".

vitorsrg: Thanks! ;D

It seams that just two tests in the while library use `assert_frame_equal` instead of a direct `assert`, so i think these shouldn't differ from the norm

PR #127, opened 2020-02-27
Startified split

gustavo-momente:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
In some cases data doesn't depend on time or space, but in their class. Therefore, to have a fair evaluation we have to split the data into train and test datasets keeping the same class ratio.

We introduce an stratified dataset  splitter that can be used during model training.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -3,3 +3,4 @@ pytest-cov>=2.6.1,<3
 pytest-xdist>=1.26.1,<2
 mypy>=0.670,<1
 codecov>=2.0,<3
+hypothesis>=5.5.4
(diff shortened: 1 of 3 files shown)


src/fklearn/resources/VERSION
```diff
@@ -1 +1 @@
-1.16.1
+1.16.2
```
arthurpaulino: Shouldn't it be 1.17.1 since there's a new feature instead of a bug fix?

gustavo-momente: Sure, just did it. Didn't know how we're doing versioning :)

src/fklearn/resources/VERSION
```diff
@@ -1 +1 @@
-1.16.1
+1.17.1
```
gnsantos: It would be version `1.17.0`, no?

PR #134, opened 2020-05-12
Add ecdf mapping to learner logs

vinipd:
### Status
**READY**

### Todo list
- [x] Tests added and passed

### Background context
In some cases we need to be able to have access to the learned ecdf map.

Adding the learned ecdf map to the learner log, as well as fixing the its prediction function log name.

### DIFF
--- a/src/fklearn/training/transformation.py
+++ b/src/fklearn/training/transformation.py
@@ -186,12 +186,16 @@ def ecdfer(df: pd.DataFrame,
 
     ecdf = ed.ECDF(values)
 
+    labels = ecdf.x
+    function_image = map(lambda image: base + sign * max_range * image, ecdf.y)
+
     def p(new_df: pd.DataFrame) -> pd.DataFrame:
         return new_df.assign(**{ecdf_column: (base + sign * max_range * ecdf(new_df[prediction_column]))})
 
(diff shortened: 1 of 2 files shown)


tests/training/test_transformation.py
```diff
@@ -524,19 +524,21 @@ def test_ecdfer():
     ecdf_column = "prediction_ecdf"
     max_range = 1000
 
-    pred_fn, data, log = ecdfer(fit_df, ascending, prediction_column, ecdf_column, max_range)
+    pred_fn, data, log = ecdfer(fit_df, ascending, prediction_column, ecdf_column, max_range, store_mapping=False)
```
gnsantos, line 527: Parametrize this test with the value of `store_mapping` and ajust the `assert` statement to:

```python
assert (log["ecdfer"].get("map") is not None) == store_mapping
```

src/fklearn/training/transformation.py
```diff
@@ -186,17 +190,23 @@ def ecdfer(df: pd.DataFrame,
 
     ecdf = ed.ECDF(values)
 
+    labels = ecdf.x
+    function_image = map(lambda image: base + sign * max_range * image, ecdf.y)
```
caique-lima, line 194: maybe we should set this only if the user passed `store_mapping = True`

PR #138, opened 2020-05-14
space_time_split_dataset slowness

jessicadesousa:
### Status
**READY**

### Background context
The function `space_time_split_dataset` take too much time when is dealing with big data samples. 

The proposal is to change the numpy filtering by use a pandas difference. This change can improve considerable the speed of the dataset creation.


sadikneipp:
Nice! Are there any measures on how faster it is? I think it would be nice to have this info in the discussion

millengustavo:
> Nice! Are there any measures on how much faster it is? I think it would be cool to have this info in the discussion

Hi @sadikneipp 

I ran some benchmark tests comparing the two versions. The results can be read in full here: [https://github.com/millengustavo/benchmark_fklearn/blob/master/benchmark_space_time_split_dataset.ipynb](https://github.com/millengustavo/benchmark_fklearn/blob/master/benchmark_space_time_split_dataset.ipynb)

**TL:DR**:
- Major speed up (~600x faster) for large datasets (1M rows) with many unique values in the `space_column`
- Pratically no difference in speed for datasets with few distinct values in the `space_column`

> "It takes about 30-60 minutes to split a 0.5-1million rows dataset, which seems somewhat excessive" - @tatasz 

- The affirmative seems to be coherent with the times I found in my computer with 1M rows and `"id": list(map(lambda x: "id%d" % x, np.random.randint(0, 1000000, n)))` -> 34min 42s
- The fix by @jessicadesousa brings the time down from 34min 42s to 3.32s

Given these preliminary results I believe it is interesting to incorporate the changes made by @jessicadesousa

PR #140, opened 2020-06-29
Add learner to return a fair calibration

caique-lima:
### Status
**IN DEVELOPMENT**

### Background context

Sometimes even if we are not using a sensitive factor as one of our model's features, we may have an uncalibrated model, where different groups from a sensitive factor have different target mean. The goal here is to generate calibrated band where any group from a sensitive factor have the same target mean.

We introduced the `find_thresholds_with_same_risk` learner, where for each group within a sensitive factor we try to find the thresholds that keep the target mean the same. This learner will add a column called `fair`(default value of `output_column`) with a band that a specific member of a group should be attributed in order to follow our metric constraints.


src/fklearn/training/calibration.py
```diff
@@ -67,3 +68,98 @@ def p(new_df: pd.DataFrame) -> pd.DataFrame:
...
+
+    Parameters
+    ----------
+
+    df : pandas.DataFrame
+        A Pandas' DataFrame with features and target columns.
+        The model will be trained to predict the target column
+        from the features.
+
+    sensitive_facor: str
```
tatasz, line 92: Typo here, facor instead of factor

tatasz:
Random thought, does it really need the ecdf is input?
Applying ECDF to classifier results is a very particular case, and not everybody does this, so maybe working on some generalisation where input doesn't have to be ecdf-ed would make this applicable to more problems.

silvanofilhonu:
Yeah. Now the code doesn't require it to be a ECDF anymore. They just have to be ordered bands.
I just need to rename the variables

silvanofilhonu (approved):
LGTM 👍

PR #141, opened 2020-07-02
Use dataframes instead of Pool in catboost learner

caique-lima:
### Status
**READY**

### Background context
During some tests we noticed that when we try to pickle a pipeline that uses the catboost learner we get the following error:

```
 File "stringsource", line 2, in _catboost._PoolBase.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
```

As pickle is a common way to serialize Fklearn models, we thought would be good to standardize this learner to enable users to pickle their pipelines.

Instead of use Pool to pass data to our trained model during the prediction time, we chose to use a dataframe, as this allow us to pickle the `p()` method

Given context why we kept the auxiliar method. Testing the `TreeExplainer` for multiclass models we got the following error:
`ValueError: could not broadcast input array from shape (x) into shape (y)`

This seems to be a well known error, as reported here:

https://github.com/slundberg/shap/issues/753
https://github.com/slundberg/shap/issues/750

- Should we add tests to all learners to guarantee we can pickle?

### DIFF
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -313,12 +313,9 @@ def catboost_classification_learner(df: pd.DataFrame,
 
     def p(new_df: pd.DataFrame, apply_shap: bool = False) -> pd.DataFrame:
 
-        dtest = Pool(new_df[features].values, feature_names=list(map(str, features)),
-                     cat_features=cat_features)
-
-        pred = cbr.predict_proba(dtest)[:, 1]
+        pred = cbr.predict_proba(new_df[features])[:, 1]
         if params["objective"] == "MultiClass":
-            pred = cbr.predict_proba(dtest)
+            pred = cbr.predict_proba(new_df[features])
             col_dict = {prediction_column + "_" + str(key): value
                         for (key, value) in enumerate(pred.T)}
             col_dict.update({prediction_column: pred.argmax(axis=1)})
@@ -328,7 +325,7 @@ def catboost_classification_learner(df: pd.DataFrame,
         if apply_shap:
             import shap
             if params["objective"] == "MultiClass":
-                shap_values = cbr.get_feature_importance(type=catboost.EFstrType.ShapValues, data=dtrain)
+                shap_values = cbr.get_feature_importance(type=catboost.EFstrType.ShapValues, data=df[features])
                 # catboost shap returns a list for each row, we reformat it to return
                 # a list for each class
                 shap_values = shap_values.transpose(1, 0, 2)
@@ -340,7 +337,7 @@ def catboost_classification_learner(df: pd.DataFrame,
 
             else:
                 explainer = shap.TreeExplainer(cbr)
-                shap_values = explainer.shap_values(dtest)
+                shap_values = explainer.shap_values(new_df[features])
                 shap_expected_value = explainer.expected_value
                 shap_values = list(shap_values)
                 shap_output = {"shap_values": shap_values,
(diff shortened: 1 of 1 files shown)


src/fklearn/training/classification.py
```diff
@@ -6,6 +6,7 @@
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.linear_model import LogisticRegression
 from sklearn import __version__ as sk_version
+from catboost import CatBoostClassifier
```
EstevaoUyra, line 9: You don't need this here, do you? You are importing inside each function. It may break the independence from catboost installation when doing such as fklearn[xgboost]

src/fklearn/training/classification.py
```diff
@@ -230,6 +231,20 @@ def p(new_df: pd.DataFrame, apply_shap: bool = False) -> pd.DataFrame:
...
+    """
+    """
+    import catboost
+    dtrain = catboost.Pool(df[features].values, df[target].values, weight=weights,
+                           feature_names=list(map(str, features)),
+                           cat_features=cat_features)
+    return cbr.get_feature_importance(type=catboost.EFstrType.ShapValues,
+                                      data=dtrain)
+
+
```
EstevaoUyra, line 247: afaik you can use shap on catboost directly from TreeExplainer, exactly as you do on lgb (L548) isn't it better to standardize?

src/fklearn/training/classification.py
```diff
@@ -313,12 +328,9 @@ def catboost_classification_learner(df: pd.DataFrame,
 
     def p(new_df: pd.DataFrame, apply_shap: bool = False) -> pd.DataFrame:
 
-        dtest = Pool(new_df[features].values, feature_names=list(map(str, features)),
-                     cat_features=cat_features)
-
-        pred = cbr.predict_proba(dtest)[:, 1]
+        pred = cbr.predict_proba(new_df[features])[:, 1]
```
EstevaoUyra, line 331: this should be inside the `else` in L325/R337. In the `if` case (MultiClass) the prediction is calculated twice, with overwriting.

EstevaoUyra, line 331: this should be inside the else in L325/R337. In the if case (MultiClass) the prediction is calculated twice, with overwriting.

EstevaoUyra: (just tried to have the full if/else in the comment but didnt manage to. The subsequent lines:
```
        pred = cbr.predict_proba(new_df[features])[:, 1]

        if params["objective"] == "MultiClass":
            pred = cbr.predict_proba(dtest)
            pred = cbr.predict_proba(new_df[features])
            col_dict = {prediction_column + "_" + str(key): value
                        for (key, value) in enumerate(pred.T)}
            col_dict.update({prediction_column: pred.argmax(axis=1)})
        else:
            col_dict = {prediction_column: pred}
```

src/fklearn/training/classification.py
caique-lima: Nice! I was looking for it. This will enable us remove this auxiliar method 🚀

caique-lima: I've updated the description with the reason to keep this auxiliar method. It's not the best, but does the job

src/fklearn/training/classification.py
```diff
@@ -230,6 +230,46 @@ def p(new_df: pd.DataFrame, apply_shap: bool = False) -> pd.DataFrame:
...
 
 
+@curry
+def _get_catboost_shap_values(df: pd.DataFrame, cbr: Any,
+                              features: List, target: str,
+                              weights: List, cat_features: List) -> np.array:
+    """
+    Auxiliar method to allow us to get shap values for Catboost multiclass models
+
+    This method was create to allow us serialize catboost models as pickle without any issues
```
vinipd, line 240:
```suggestion
    This method exists to allow us to serialize catboost models as pickle without any issues
```

PR #146, opened 2020-08-06
Flexibilize dependencies, as of 2020-07

vitorsrg:
### Status
**READY**

### Background context
Some of the bounds are more than a year old already.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-cloudpickle>=0.8.0,<0.9.0
-joblib>=0.13.2,<0.14.0
-numpy>=1.16.4,<1.17.0
-pandas>=0.24.1,<0.25
-scikit-learn>=0.21.2,<0.22.0
+cloudpickle>=0.8.0,<1.6.0
+joblib>=0.13.2,<0.17.0
+numba>=0.48.0,<0.50
+numpy>=1.16.4,<1.20.0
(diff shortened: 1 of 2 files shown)


caique-lima:
Just saw that parso was changed to keep compatibility with jedi

caique-lima:
Regarding numba https://github.com/nubank/fklearn/pull/139

> numba to keep compatibility with the previous api with numba.errors module

I think with previous requirements some of them used numba.errors, probably fixed with more flexible requirements

vitorsrg:
The plan was to allow newer versions of numpy, pandas and sklearn.

For cloudpickle and joblib, I guess I'll change them back, then.

I didn't remove numba, but it seams I I accidentally sorted the file somehow

vitorsrg:
> Do we have a reason to change cloudpickle? historically it's been an issue to deal with different versions.

Doesn't bumping minor justify breaking pickles from previous versions?

caique-lima:
> Doesn't bumping minor justify breaking pickles from previous versions?

Sadly, yes. As we still uses the 0 major versions, things break all the time.

We can create a milestone for Fklearn 2 and put cloudpickle 1 as one of the tasks, this will probably make everything more stable. We also can start to support only Pandas >= 1, <2. We can take a look at things we should deprecate, and break and put everything in the same release

PR #149, opened 2020-08-15
Include min and max values in the extremeintervals on quantile biner.

### LINKED ISSUE #112: quantile_biner
### Instructions
Quantile biner creates an extra class either for `min` or for `max`.

### Code sample
```python
test = pd.DataFrame({
    "id": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
    "value": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
})

transformation.quantile_biner(test, columns_to_bin=['value'], q=2, right=False)[1]
```

### Problem description
This will result in 3 classes, with the  `min` value in a separate class if `right=True`, and max value in a separate class otherwise.

If fact, it creates an extra class for min or max for all the values of q (both int and list of quantiles).

### Expected behavior
Behave like pd.qcut, when setting q=n will create exactly n classes.

### Possible solutions
include min and max values in the extreme intervals, regardless the `right` setting.

gabrieldi95:
### Status
**READY**

- [ ] Issue: closes #112

### Background context
quantile_biner would result in n+1 quantiles when the min or max values were equal to either the first or last quantile.

Change values categorized as the '0' quantile to 1 and values categorizes as 'n+1' (n being the wanted number of quantiles)
to n.

### DIFF
--- a/src/fklearn/training/transformation.py
+++ b/src/fklearn/training/transformation.py
@@ -643,9 +643,12 @@ def quantile_biner(df: pd.DataFrame,
     bin_getter = lambda col: pd.qcut(df[col], q, retbins=True)[1]
     bins = {column: bin_getter(column) for column in columns_to_bin}
 
+    def _fix_values(values, max):
+        return [1.0 if value == 0 else (value - 1.0 if value == max else value) for value in values]
+
     def p(new_df: pd.DataFrame) -> pd.DataFrame:
         col_biner = lambda col: np.where(new_df[col].isnull(), nan, np.digitize(new_df[col], bins[col], right=right))
-        bined_columns = {col: col_biner(col) for col in columns_to_bin}
+        bined_columns = {col: _fix_values(col_biner(col), len(bins[col])) for col in columns_to_bin}
         return new_df.assign(**bined_columns)
 
     p.__doc__ = learner_pred_fn_docstring("quantile_biner")


gabrieldi95:
Tests failed probably because of already existing errors... File tested is ok.
image

src/fklearn/training/transformation.py
```diff
@@ -643,9 +643,12 @@ def quantile_biner(df: pd.DataFrame,
     bin_getter = lambda col: pd.qcut(df[col], q, retbins=True)[1]
     bins = {column: bin_getter(column) for column in columns_to_bin}
 
+    def _fix_values(values, max):
+        return [1.0 if value == 0 else (value - 1.0 if value == max else value) for value in values]
+
     def p(new_df: pd.DataFrame) -> pd.DataFrame:
         col_biner = lambda col: np.where(new_df[col].isnull(), nan, np.digitize(new_df[col], bins[col], right=right))
-        bined_columns = {col: col_biner(col) for col in columns_to_bin}
+        bined_columns = {col: _fix_values(col_biner(col), len(bins[col])) for col in columns_to_bin}
```
vitorsrg, line 651: i think `np.clip(col_biner(col), 0, len(bins[col]))` could be more idiomatic here

vitorsrg:
the mock test seams to fail because this PR changes the function behaviour...

maybe this change could be triggered by some parameter in a backwardly compatible way

caique-lima:
@gabrieldi95 will you keep working on this? Seems that the tests are broken. Let me know, otherwise we will close this PR for now, later you can feel free to reopen it.

caique-lima:
No updates in the last few months

PR #150, opened 2020-08-16
Ohe drop_first_column parameter fixed

### LINKED ISSUE #148: One-hot encoder's drop_first_column parameter does nothing
The parameter is not used in the transformation code.
https://github.com/nubank/fklearn/blob/3f3226df3489a63e599537d1b5413779a0614836/src/fklearn/training/transformation.py#L668

jessicadesousa:
### Status
**READY**

### Todo list
- [x] Tests added and passed
- [x] Issue: closes #148 

### Background context
The parameter `drop_first_column` in the onehot_categorizer was unused, so it was fixed.


tests/training/test_transformation.py
```diff
@@ -360,35 +369,38 @@ def test_onehot_categorizer():
         ("fklearn_feat__region==SP", [0, 0, 1, 0])
     )))
 
-    # Test without hardcoding NaNs
+    expected_output_test_drop_first_hardcode = pd.DataFrame(OrderedDict((
```
gabrielferreira95, line 372: Would it make sense to add `drop_first`+`no_hardcode` tests? 
Cases like: `expected_output_train_drop_first_no_hardcode` and `expected_output_test_drop_first_no_hardcode`.

jessicadesousa: It has the four combinations:
Line 381: Without hardcoding NaNs & not dropping first column
Line 392:       with hardcoding NaNs & not dropping first column
Line 403: Without hardcoding NaNs & dropping first column
Line 415:       with hardcoding NaNs & dropping first column

gabrielferreira95: Yeah, you're right :) They are there. 
I just would expect this case `expected_output_train_drop_first`, for instance, to be named `expected_output_train_drop_first_no_hardcode`. I think it would be more clear.

gabrielferreira95: I had not realized those cases were there because I was searching for `drop_first_no_hardcode`. I thought it made more sense, considering the other cases names. What do you think?

gabrielferreira95 (approved):

caique-lima (approved):

PR #152, opened 2020-09-28
🐛 [Bugfix] split_evaluator_extractor now accepts eval_name

### LINKED ISSUE #151: split_evaluator_extractor not fully compliant with split_evaluator
### Problem description
`split_evaluator` has an optional parameter called `eval_name`, with a default value of `None`. If this parameter is used, `split_evaluator_extractor`, which is supposed to facilitate the extraction of information in the logs generated by `split_evaluator`, does not work. It only works for the default case.

### Code sample
```python
import pandas as pd
import numpy as np
from sklearn.datasets import load_breast_cancer
from fklearn.validation.evaluators import split_evaluator, roc_auc_evaluator
from fklearn.metrics.pd_extractors import split_evaluator_extractor, evaluator_extractor

loaded_data = load_breast_cancer()
size = len(loaded_data['target'])
dataset = pd.DataFrame(loaded_data['data']).assign(
  target=loaded_data['target'], 
  prediction=np.random.rand(1, size)[0],
  split=np.random.randint(2, size=size)
)

eval_fn = roc_auc_evaluator(target_column='target', eval_name='roc_auc')
base_extractor = evaluator_extractor(evaluator_name='roc_auc')

# Default case
split_fn_default = split_evaluator(eval_fn=eval_fn, split_col="split")
default_logs = split_fn_default(dataset)
default_metrics = split_evaluator_extractor(default_logs, split_col="split", base_extractor=base_extractor, split_values=[0,1])
print(default_metrics)

#    roc_auc  split_evaluator__split
#0  0.488234                       0
#0  0.448012                       1

# Bug case - with eval_name
split_fn_named = split_evaluator(eval_fn=eval_fn, split_col="split", eval_name="named_eval")
named_logs = split_fn_named(dataset)
named_metrics = split_evaluator_extractor(named_logs, split_col="split", base_extractor=base_extractor, split_values=[0,1])
print(named_metrics)

#   roc_auc  split_evaluator__split
#0      NaN                       0
#0      NaN                       1
```

### Expected behavior
`split_evaluator_extractor` should be able to extract logs generated by `split_evaluator` with an `eval_name`.

### Possible solutions
Include a `eval_name` parameter on `split_evaluator_extractor` , just like `temporal_split_evaluator_extractor` has.

mlikoga:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes #151 

Added optional parameter `eval_name` to `split_evaluator_extractor`

`split_evaluator_extractor` param `split_col` should be optional, as it is irrelevant if `eval_name` is used, but I do not know how to do this without breaking the API


src/fklearn/metrics/pd_extractors.py
```diff
@@ -19,16 +19,19 @@ def combined_evaluator_extractor(result, base_extractors):
 
 
 @curry
-def split_evaluator_extractor_iteration(split_value, result, split_col, base_extractor):
-    key = 'split_evaluator__' + split_col + '_' + str(split_value)
+def split_evaluator_extractor_iteration(split_value, result, eval_name, base_extractor):
+    key = eval_name + '_' + str(split_value)
     return (base_extractor(result.get(key, {}))
-            .assign(**{'split_evaluator__' + split_col: split_value}))
+            .assign(**{eval_name: split_value}))
```
caique-lima, line 25: I don't think we should break the api(as you pointed in the description) to fix this bug, I will try to think about ways to solve it 🤔 

I think we are missing tests for this, so this is the reason why everything succeeded

caique-lima (changes requested):

src/fklearn/metrics/pd_extractors.py
mlikoga: hmmm I was split_evaluator_extractor_iteration as a private function

tests/metrics/test_pd_extractors.py
```diff
@@ -19,6 +20,51 @@
 from fklearn.validation.validator import validator
 
 
+def test__split_evaluator_extractor_iteration__default():
+    logs = {
+        'split_evaluator__split_0': {'roc_auc': 0.48},
+        'split_evaluator__split_1': {'roc_auc': 0.52},
+    }
```
caique-lima, line 27: As this `logs` var is being used in more than one test would be cool to transform it into a fixture instead of repeat the definition, same for the named logs in `test__split_evaluator_extractor_iteration__with_eval_name` test

caique-lima: Also most of these tests are very similar, would be nice to use the pytest parametrize to just change the parameters. [Example](https://github.com/nubank/fklearn/blob/master/tests/training/test_pipeline.py#L14)

caique-lima (approved):
No blockers, just some small suggestions on how to improve the tests. Let me know if you will do the proposed changes, otherwise we can merge this one

tests/metrics/test_pd_extractors.py
mlikoga: done

caique-lima (approved):
nice job with test refactoring 👏

PR #158, opened 2021-01-24
Fix Spatial LC check

robotenique:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context

Spatial LC wasn't verifying correctly if the upper value of the percentage param was bigger than (as it should - you can't select more than 100% of the IDs). This PR just fix this check.

Fix the check in Spatial LC and add test.


gcbeltramini (approved):

gcbeltramini (approved):
Nice catch, and thanks for adding a test that captures the behavior.

PR #160, opened 2021-02-02
Add decorator to value mapper

tatasz:
### Status
**IN DEVELOPMENT**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Value mapper transformation currently overwrites the values in a column. Update the `column_duplicatable` decorator to allow storing both mapped and original values.

### DIFF
--- a/src/fklearn/preprocessing/schema.py
+++ b/src/fklearn/preprocessing/schema.py
@@ -117,8 +117,6 @@ def column_duplicatable(columns_to_bind: str) -> Callable:
                     columns_to_duplicate: Any = (kwargs[columns_to_bind] if columns_to_bind in
                                                  kwargs.keys() else
                                                  args[child_spec.args.index(columns_to_bind)])
-                    if isinstance(columns_to_duplicate, dict):
-                        columns_to_duplicate = list(columns_to_duplicate.keys())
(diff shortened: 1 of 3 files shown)


src/fklearn/preprocessing/schema.py
```diff
@@ -117,6 +117,8 @@ def _learn(df: pd.DataFrame) -> LearnerReturnType:
                     columns_to_duplicate: Any = (kwargs[columns_to_bind] if columns_to_bind in
                                                  kwargs.keys() else
                                                  args[child_spec.args.index(columns_to_bind)])
+                    if isinstance(columns_to_duplicate, dict):
```
caique-lima, line 120: What is the expected behaviour with this if statement?

Seems that the method `feature_duplicator` already solves the case if columns_to_duplicate is a dict(L53)

I've tested both with and without it and got the same results, do you think it is needed?

tatasz: Hmmmmmm it was failing here, but maybe I was testing on an older version, will check this out :thanks:

tatasz: Yeah it works, probably i'm just a dummy

caique-lima (approved):

PR #162, opened 2021-02-09
Update setuptools in CI scripts

caique-lima:
### Status
**READY**

### Background context

In our last try to release the package we had some problems installing `cryptography` package, usually we can solve these issues keeping setuptools updated

### DIFF
--- a/scripts/create_venv.sh
+++ b/scripts/create_venv.sh
@@ -9,4 +9,4 @@ source $cur_dir/helpers.sh
 
 activate_venv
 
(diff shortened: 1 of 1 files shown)


gcbeltramini (approved):
The [current version is 53.0.0](https://pypi.org/project/setuptools/#history), i.e., two possible breaking changes. The only problems I remember having with `setuptools` is having to bump it, and not having something breaking because of a major release. So I think this is good.

PR #165, opened 2021-04-08
Update transformation.py

tatasz:
# Fix mistake in docstrings


gcbeltramini (approved):
I checked the tests in https://github.com/nubank/fklearn/blob/a3de09a6949089f47633e64b536a808650d7e9d1/tests/training/test_transformation.py, and your suggestion seems correct. Just a side note, the indentation in that test file makes is harder to understand what's going on.

PR #169, opened 2021-10-19
Causal val Funcs

matheusfacure:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Add common causal evaluation techniques, such as cumulative effect curve and cumulative gain curve

First, this PR adds a list of possible metrics that one can use as effects.
Then, it uses those effects for causal model validation curves.
Finally, it adds functions to compute the area under those causal inference curves

1. Effects
2. Curves
3. AUC

Need to integrate the causal metrics with standard fklearn validators. This should be easy to do, as the causal AUC function all return a float.

### DIFF
--- a/src/fklearn/causal/effects.py
+++ b/src/fklearn/causal/effects.py
@@ -1,16 +1,16 @@
 import pandas as pd
 from toolz import curry
+from typing import Dict, Callable
 
 from fklearn.validation.evaluators import (spearman_evaluator, correlation_evaluator, linear_coefficient_evaluator,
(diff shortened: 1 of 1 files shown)


arthurpaulino:
> @arthurpaulino, I'm not sure. I could implement everything as evaluators and use them instead. Feels a bit odd, though, since the purpose is very different.

A possible solution is to implement the new ones (`linear_effect` and `linear_effect_ci`) as evaluators and then make interfaces reuse those implementations inside the `causal` module, but without the need to make new tests for the spearman and pearson effects since their evaluators are already tested

PR #174, opened 2021-11-01
Add NonParametric Double/Debias Machine Learning

matheusfacure:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Implements the Non Parametric version of Double Machine Learning, according to [Foster (2019)](https://arxiv.org/abs/1901.09036). This is a generic ML model to predict heterogeneous treatment effects: E[delta T | X]

* Introduces a cate (Conditional Average Treatment Effect Learner) function. The idea is for causal inference learner to have a similar signature to fklearn learners. The only difference is the addition of the treatment column. Hence, the standard signature should be 
```python
def causal_learner(df: pd.DataFrame,
                   feature_columns: List[str],
                   treatment_column: str,
                   outcome_column: str,
                   ...) -> LearnerReturnType
```

* Implements Non Parametric Double/ML
This is the most generic of the Double/ML algorithms. It takes three generic ML regressor to fit a model for treatment effect estimation.


tatasz (approved):
I wonder if an example notebook at this point would be useful.

matheusfacure:
@tatasz, soon!

src/fklearn/causal/cate_learning/double_machine_learning.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@curry
+@log_learner_time(learner_name='non_parametric_double_ml_learner')
+def non_parametric_double_ml_learner(df: pd.DataFrame,
+                                     feature_columns: List[str],
+                                     treatment_column: str,
+                                     outcome_column: str,
+                                     debias_model: RegressorMixin = GradientBoostingRegressor(),
+                                     debias_feature_columns: List[str] = None,
```
dieggoluis, line 41: Mutable default arguments must be avoided in python. Please use `None` instead, then 
```
if debias_model is None:
   debias_model = GradientBoostingRegressor()
```

dieggoluis (changes requested):

src/fklearn/causal/cate_learning/double_machine_learning.py
dieggoluis: https://docs.python-guide.org/writing/gotchas/

matheusfacure: Nice to know!

dieggoluis (approved):

PR #176, opened 2021-11-17
Fix curry issue

### LINKED ISSUE #175: Double/Debiased Resusing Models when Curried
### Code sample
A common pattern in fklearn is to partially define a training function to create a `fit_fn` that will later be reused. For example, let's say we want to do feature selection. We can create a `fit_fn` passing all the parameters to a learner but the `feature_columns`

```python
fit_fn = non_parametric_double_ml_learner(
    df=train_set_eng.head(10000),
    treatment_column = treatment,
    outcome_column = outcome,
    debias_model = LGBMRegressor(),
    denoise_model = LGBMRegressor(),
    final_model = LGBMRegressor(),
    prediction_column= "dml_elast_prediction",
)
```
Then, we can try estimating the above model with different feature sets. For example, we can fit a model with only 3  (`m1`) features and another one with 6 features (`m2`). 

```python
f1 = features[:3]
f2 = features[:6]

m1 = fit_fn(feature_columns=f1)[0]
m2 = fit_fn(feature_columns=f2)[0]
````

However, once we try to make predictions with the first model, we get the following error

```python
m1(train_set_eng[f1].head())["dml_elast_prediction"]
```
```
ValueError: Number of features of the model must match the input. Model n_features_ is 6 and input n_features is 3 
```

So, `m1` is requiring 6 features? But we only trained it with 3? What is happening?

### Problem description
The problem is that fklearn is reusing the same instance of the model parameters in the learner function

```python
debias_model = LGBMRegressor(),
denoise_model = LGBMRegressor(),
final_model = LGBMRegressor(),
```

That way, since we train `m2` after `m1` and we reuse instance across both models, `m1` ends up having the model instance from `m2`, which requires 6 features. 

### Possible solutions
One easy solution is to use `from sklearn.base import clone` in the learner [here](https://github.com/nubank/fklearn/blob/ca1d8bd5b0667a5b9076fe02cc840f31e52b4ae1/src/fklearn/causal/cate_learning/double_machine_learning.py#L117)

```python
debias_model = GradientBoostingRegressor() if not debias_model else clone(debias_model)
denoise_model = GradientBoostingRegressor() if not denoise_model else clone(denoise_model)
final_model = GradientBoostingRegressor() if not final_model else clone(final_model)
```

That ensures we are always creating a new instance when calling this learner function. 

This is probably happening here too 
https://github.com/nubank/fklearn/blob/ca1d8bd5b0667a5b9076fe02cc840f31e52b4ae1/src/fklearn/causal/cate_learning/double_machine_learning.py#L29

matheusfacure:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes #175 

### Background context
Fix the issue above with the solution proposed there. Also, add tests to check that behavior.

### DIFF
--- a/src/fklearn/causal/cate_learning/double_machine_learning.py
+++ b/src/fklearn/causal/cate_learning/double_machine_learning.py
@@ -6,6 +6,7 @@ from sklearn import __version__ as sk_version
 from sklearn.base import RegressorMixin
 from sklearn.model_selection import KFold
 from sklearn.ensemble import GradientBoostingRegressor
+from sklearn.base import clone
 from toolz import curry
 from typing import Union
 
@@ -25,7 +26,7 @@ def _cv_estimate(model: RegressorMixin,
     cv_pred = pd.Series(np.nan, index=train_data.index)
 
     for train, test in cv.split(train_data):
-        m = model.fit(train_data[features].iloc[train], train_data[y].iloc[train])
+        m = clone(model, safe=False).fit(train_data[features].iloc[train], train_data[y].iloc[train])
         cv_pred.iloc[test] = m.predict(train_data[features].iloc[test])
         models += [m]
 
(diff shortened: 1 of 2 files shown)


src/fklearn/causal/cate_learning/double_machine_learning.py
```diff
@@ -151,7 +152,7 @@ def p(new_df: pd.DataFrame) -> pd.DataFrame:
         'prediction_column': prediction_column,
         'package': "sklearn",
         'package_version': sk_version,
-        'feature_importance': dict(zip(features, model_final_fitted.feature_importances_)),
+        'feature_importance': None,
```
matheusfacure, line 155: There isn't a standard in sklearn models for a feature importance attribute, which causes error depending on the model you use as the final model.

tatasz (approved):

PR #178, opened 2021-12-02
Add causal curves summary

pedroig:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

Adding a function to summarize causal curves in a pandas dataframe.

Motivation:
* The current causal curve functions return a numpy array so only the y-axis of the curves is returned. Therefore, in order to plot the curves, it's necessary to rebuild an inner logic to construct the x-axis.
* We often need to plot all the causal curves to analyze and debug results. However, if we use the individual curve functions in each of those cases, we would have a meaningful redundancy in the computation since the core computation from the cumulative effect curve would be executed multiple times.

### DIFF
--- a/src/fklearn/causal/validation/curves.py
+++ b/src/fklearn/causal/validation/curves.py
@@ -220,7 +220,7 @@ def effect_curves(
      Creates a dataset summarizing the effect curves: cumulative effect, cumulative gain and
      relative cumulative gain. The dataset also contains two columns referencing the data
      used to compute the curves at each step: number of samples and fraction of samples used.
-     Moreover one column indicating the cumulative gain for a corresponding random model is 
+     Moreover one column indicating the cumulative gain for a corresponding random model is
      also included as a benchmark.
 
      Parameters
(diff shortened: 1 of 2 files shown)


src/fklearn/causal/validation/curves.py
```diff
@@ -202,3 +204,75 @@ def relative_cumulative_gain_curve(df: pd.DataFrame,
...
+     Returns
+     ----------
+     summary curves dataset: pd.DataFrame
+         The dataset with the results for multiple validation causal curves according to the predictions ordering.
+    """
+
+    size: int = df.shape[0]
+    n_rows: List[int] = list(range(min_rows, size, size // steps)) + [size]
+
+    cum_effect: np.ndarray = cumulative_effect_curve(
```
matheusfacure, line 260: Should't this function be genertic? I mean, why only support the cumm effect and not other curves?

pedroig: hmmm, what other curves? Could you give an example? I was thinking about supporting only the three curves that we have currently: the cumulative effect as shown here and the cumulative gain and the relative cumulative gain that are built on top of the cumulative effect curve.

matheusfacure: Humm, I get it know. You are already using everything.

matheusfacure (approved):
Looks good.

PR #182, opened 2021-12-30
Run tests with Python 3.9

gcbeltramini:
### Status

**READY**

### Background context

Some libraries are dropping support to Python 3.6, [such as pip](https://pip.pypa.io/en/stable/news/#v22-0), so we need to prepare to support later versions of Python.

Add test in CircleCI to run in an image with Python 3.9. A new version doesn't have to be released, because only test files were modified.

File `.circleci/config.yml` then the Python files.

The step "Install package" takes about 10 minutes to run in Python 3.9, probably due to `scikit-learn` (in Python 3.6 they take ~1 minute); and the unit tests take ~15 minutes (in Python 3.6 they take ~1 minute).


tatasz (approved):
We go for it and if something breaks later on we fix it.

PR #184, opened 2022-01-19
Minor fixes on the documentation

erickisos:
Some of the links were not currently active (mainly on github side).
There was a link with a minor error on the underscore position.
Added some extra context for some sentences.

### Status

**READY**

### Todo list
- [x] Documentation

### Background context

Just added some fixes that I encountered when I tried to start contributing.

As a side note, if you see something weird with the text or some syntactic problems, just let me know and we can add it to this PR.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -5,25 +5,30 @@ Getting started
 Installation
 ------------
 
-The fklearn library is Python 3.6 compatible only. In order to install it using pip, run::
+The fklearn library is Python 3.6 compatible only. In order to install it using pip, run:
+
+.. code-block:: bash
 
     pip install fklearn
 
+You can also install from the source:
 
-You can also install from the source::
+.. code-block:: bash
 
     # clone the repository
-    $ git clone -b master https://github.com/nubank/fklearn.git --depth=1
+    git clone -b master https://github.com/nubank/fklearn.git --depth=1
     
     # open the folder
-    $ cd fklearn
+    cd fklearn
     
     # install the dependencies
-    $ pip install -e .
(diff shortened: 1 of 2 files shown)


arthurpaulino (dismissed):
Thanks!!

docs/source/getting_started.rst
```diff
@@ -45,13 +51,15 @@ The **transformed training data** is usually just the prediction function applie
...
+Learner functions are usually partially initialized (curried) before being passed to pipelines or applied to data:
+
+.. code-block:: python
 
     from fklearn.training.regression import linear_regression_learner
     from fklearn.training.transformation import capper, floorer, prediction_ranger
 
     # initialize several learner functions
-    capper_fn = capper(columns_to_cap=["income"], precomputed_caps={"income": 50,000})
+    capper_fn = capper(columns_to_cap=["income"], precomputed_caps={"income": 50_000})
```
gcbeltramini, line 62: Shouldn't this be?

```suggestion
    capper_fn = capper(columns_to_cap=["income"], precomputed_caps={"income": 50000})
```

erickisos: afaik [both are valid options](https://peps.python.org/pep-0515/), it is better with your suggestion, but I didn't feel confident enough to remove the separator.
If you think it is readable without the separator, it's ok for me

gcbeltramini: I prefer without `_`, because that syntax is not usual.

erickisos: That's a good point, I will add your suggestion then @gcbeltramini

gcbeltramini (approved):

Issue #185, opened 2022-01-19
Migration plan for python 3.8, 3.9 or some "newer tooling"


erickisos:
Hey everyone!

I recently started trying to contribute to this project and I saw PR #182, so I just want to know which would be the expected support level for that version or future versions of the language and whether would be great to start helping with something like that.

Also, which kind of features/tooling would be nice to have in here? And which ones seems valuable for the project itself, I mean, things like the following:

- pyproject.toml file for centralized configurations (this includes mypy, flake8, isort, or maybe black(?) in the same config file).
- Poetry as the default deps and venvs manager(?)
- Github actions instead of CircleCI sounds viable? or in case that doesn't make sense at all, and in that case, do you have currently a code quality workflow checking all the points, linting, formatting, code complexity, docstrings, and typing? Are you open to considering options to avoid using custom shell scripts?

Sorry if this is dumb or something, but I saw the gitter room a little inactive and I don't know if some of those points can be a real deal or if I'm saying things that other persons had proposed.

Thanks in advance folks 💜

caique-lima:
Hi Erick!

Since the gitter don't have a lot of action anymore, I think the best path is actually open issues here in our Repo.

I would say that anything that improves the developers experience is a nice contribution. Personally I'm really open to things like the ones you proposed if they still allow people to easily install the library using a `pip install fklearn` and don't require them to change a lot the dev environment.

> Github actions instead of CircleCI sounds viable? or in case that doesn't make sense at all, and in that case, do you have currently a code quality workflow checking all the points, linting, formatting, code complexity, docstrings, and typing? Are you open to considering options to avoid using custom shell scripts?

About this last one I would say that it is worth the investment only if Github actions brings something that CircleCI doesn't support as of today. Talking about what we provide, we do have codecoverage, lint, formatting and typing (most of them defined [here](https://github.com/nubank/fklearn/tree/master/scripts)). And the same thing I said about the Github actions can be applied here, I wouldn't invest time to have exactly the same experience using other tools, but if it is something that can improve the experience we can plan to migrate some stuff.

Feel free to propose something related to `pyproject.toml` and/or `Poetry` and we can keep discussing. I would be happy to review the PRs and guide you on how we think about the project.

caique-lima:
I forgot to talk about different Python versions, I would be also really happy to improve the release/test flow for different versions. Today it is kinda slow, I think in the past we tried to use tox, but I'm not sure if it is the best tool

erickisos:
[...]

Issue #157, opened 2021-01-04
Dependency management


matheusfelipeog:
Hi ;)

This project uses requirements.txt as a dependency manager file. Wouldn't it be useful to use **Pipenv + Pipfile** or **Poetry + Pyproject.toml** as a substitute?

With any of the indicated you can manage all dependencies in 1 file and with features like locking dependencies with hash and the like...

---

More info:

**Pipenv:** https://pipenv.pypa.io/en/latest/
**Pipfile:** https://github.com/pypa/pipfile

**Poetry:** https://python-poetry.org/
**Pyproject.toml:** https://python-poetry.org/docs/pyproject/

dieggoluis:
Pipenv was almost a dead project for a long time (https://github.com/pypa/packaging.python.org/issues/701) and I've seen some people having quite a few issues with it. 

I've been using poetry for about six months and it looks promising. The only caveat from my experience is that it takes more time to generate the lock file and you will need to generate the equivalent requirements file if you do not want to install poetry in your docker images or servers. If you do want to use poetry with docker (https://stackoverflow.com/questions/53835198/integrating-python-poetry-with-docker). 

Pip-tools can also generate hashed dependencies, but I believe they chose not to use it. Poetry looks promising and would definitely replace some scripts of this project, but if maintainers are happy the way it is, I don't see any reason to change. I'm curious though if there is any other reason to stick with pip-tools. 🤔 @caique-lima

matheusfelipeog:
> Pipenv was almost a dead project for a long time (pypa/packaging.python.org#701) and I've seen some people having quite a few issues with it.

This is really a point, but apparently the project is back on track, according to the [release history](https://github.com/pypa/pipenv/releases). This don't deny having extra attention with this tool.

> you will need to generate the equivalent requirements file if you do not want to install poetry in your docker images or servers

It looks like Pypa will create a new command line option for [pip to support Pipfile format](https://github.com/pypa/pipfile#pip-integration-eventual).

I didn't find much similar information for pyproject.toml format. The closest discussion was this: https://github.com/pypa/pip/issues/8049

But, because both follow the [toml](https://github.com/toml-lang/toml) standard, I believe support for both files will be implemented in the pip in the near future.

gcbeltramini:
There are indeed many ways to manage dependencies in Python. We tested `Pipfile` and `poetry`. Both didn't seem to add any special advantage and their usage was not widespread. We wanted to:

[...]

PR #186, opened 2022-01-24
Poetry migration example

erickisos:
### Status

**WIP**

This is a major PR to show which would be the full change of the project if we decide to migrate everything to Poetry, we would need to add some extra changes to documentation to keep everything synced, however the major advantage is the centralized file for tool configuration and dependency management, also and just as a reference, using this would lead us eventually to having a more reliable system for multiple python versions.

The main difference is on the side of config files, for example, now the `pyproject.toml` is the main source of configuration for the majority of tools (let's say mypy, pytest), we have included the files for Github Actions because of the problem with CircleCI, and you can check the new workflows on `.github/workflows`.

TBD(?)

Relates to:
- #185
- #157
- #187


erickisos:
Now we have the full pipeline running with Poetry and Github Actions [here](https://github.com/erickisos/fklearn/actions/runs/1797293900)

PR #187, opened 2022-01-24
Github actions migration

erickisos:
### Status

**DONE**

### Background context

Recently, the CircleCI process has stopped working, so now we need to evaluate alternatives to run our existing validations, this PR is intended to solve that problem by providing the basic tasks that we already had directly on Github Actions.

Relates to #185


scripts/create_venv.sh
```diff
@@ -5,7 +5,7 @@ set -e
 cur_dir=$(dirname ${BASH_SOURCE[0]})
 source $cur_dir/helpers.sh
 
-/usr/local/bin/python3 -m venv venv
+python3 -m venv venv
```
erickisos, line 8: We have the option to keep this as 
```suggestion
/usr/bin/env python3 -m venv venv
```
But I think it would be valuable to if we add a dep on `pyenv` in order to have the expected python version, however maybe this library should be agnostic about the python version (^3.6)

erickisos:
Just in case, the checks for this PR are running directly on Github Actions on my side, but after the merge is expected to have everything running here at the main repository.

You can check the results and steps [here](https://github.com/erickisos/fklearn/actions/runs/1814315063).

erickisos:
Update: Now we have the python3.9 version of the tests running [here](https://github.com/erickisos/fklearn/actions/runs/2023538425) @caique-lima

However, maybe we can improve this in another PR by adding tox with multiple targets(?).

prcastro (approved):
LGTM

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Restoring cache
+        uses: actions/cache@v2
+        with:
+          path: ./venv
+          key: ${{ runner.os }}-python-${{ hashFiles('requirements.txt') }}
+      - name: Create virtual environment
+        run: . scripts/create_venv.sh
+      - name: Init .pypirc
+        env:
+          PYPI_USER: ${{ secrets.PYPI_USER }}
```
prcastro, line 54: Where do we set these secrets?

fpingas:
Closing as it was solved in https://github.com/nubank/fklearn/pull/227

PR #195, opened 2022-04-22
Optionally returning training scores in the validator functions

diegoarri91:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Computing training scores (metrics evaluated on the training set) can be useful to assess underfitting/overfitting and guide model selection (in e.g. the context of hyperparameter tuning). Sci-kit learn validation and hyperparameter exploration functions (such as `sklearn.model_selection` `cross_validate`, `GridSearchCV` and `RandomizedSearchCV`) provide the option of returning the training scores. This is implemented using the boolean optional parameter `return_train_score` which is False by default, meaning no training scores are returned under the default behavior. From scikit-learn documentation on this parameter: 

`return_train_score` bool, default=False
Whether to include train scores. Computing training scores is used to get insights on how different parameter settings impact the overfitting/underfitting trade-off. However computing the scores on the training set can be computationally expensive and is not strictly required to select the parameters that yield the best generalization performance.

In this PR I aim to replicate this behavior for the `validator_iteration`, `validator` and `parallel_validator` functions in fklearn/validation/validator.py file. 

In the `validator_iteration` function (this one currently trains on train and evaluates on test without evaluating on train) I added the boolean optional parameter `return_train_score` which is False by default, preserving current functionality if not used. 
- If `return_train_score=False` (default) the training metrics are not computed and the returned `logs` are exactly the same as before.
- If `return_train_score=True`, the `eval_fn` is applied also on the `train_data` and the results can be found in the returned `logs` (in `logs['train_log']['eval_results']`). This involved adding two lines of code (an `if` and the evaluation)

In the `parallel_validator_iteration` function the only changes are adding the new `return_train_score` parameter (optional, defaults to False preserving current functionality) and passing it to the `validator_iteration` function.

In the `parallel_validator` function (this one performs the train-evaluation sequence on multiple fold/splits in parallel) the only proposed changes involve adding the new `return_train_score` parameter (optional, defaults to False preserving current functionality) and passing it to the `parallel_validator_iteration` function.

In the `validator` function (this one performs the train-evaluation sequence on multiple fold/splits sequentially) the proposed changes involve adding the new `return_train_score` parameter (optional, defaults to False preserving current functionality) and passing it to the `validator_iteration` function. Additionally, this one current functionality returns only the train log corresponding to the first split (as opposed to `parallel_validator` which was already returning all). Thus, I added another `return_all_train_logs` parameter (optional, defaults to False preserving current functionality). If True returns all logs, if False preserves current functionality.

All changes so far are in the fklearn/validation/validator.py file

### DIFF
--- a/src/fklearn/validation/validator.py
+++ b/src/fklearn/validation/validator.py
@@ -21,7 +21,7 @@ def validator_iteration(data: pd.DataFrame,
                         train_fn: LearnerFnType,
                         eval_fn: EvalFnType,
                         predict_oof: bool = False,
-                        return_train_score : bool = False,
+                        return_train_score: bool = False,
                         verbose: bool = False) -> LogType:
     """
     Perform an iteration of train test split, training and evaluation.
--- a/src/fklearn/validation/validator.py
+++ b/src/fklearn/validation/validator.py
@@ -21,7 +21,7 @@ def validator_iteration(data: pd.DataFrame,
                         train_fn: LearnerFnType,
                         eval_fn: EvalFnType,
                         predict_oof: bool = False,
-                        return_train_score: bool = False,
+                        return_eval_fn_on_train: bool = False,
                         verbose: bool = False) -> LogType:
     """
     Perform an iteration of train test split, training and evaluation.
@@ -52,8 +52,8 @@ def validator_iteration(data: pd.DataFrame,
     predict_oof : bool
         Whether to return out of fold predictions on the logs
 
-    return_train_score : bool
-        Whether to include train scores
+    return_eval_fn_on_train : bool
+        Whether to apply eval_fn to the training set and return the evaluation logs in the train log
 
     Returns
     ----------
@@ -67,7 +67,7 @@ def validator_iteration(data: pd.DataFrame,
     warnings.warn(empty_set_warn) if train_data.shape[0] == 0 else None  # type: ignore
 
     predict_fn, train_out, train_log = train_fn(train_data)
-    if return_train_score:
+    if return_eval_fn_on_train:
         train_log = assoc(train_log, "eval_results", eval_fn(train_out))
 
     eval_results = []
@@ -96,7 +96,7 @@ def validator(train_data: pd.DataFrame,
               perturb_fn_train: PerturbFnType = identity,
               perturb_fn_test: PerturbFnType = identity,
               predict_oof: bool = False,
-              return_train_score: bool = False,
+              return_eval_fn_on_train: bool = False,
               return_all_train_logs: bool = False,
               verbose: bool = False) -> ValidatorReturnType:
     """
@@ -135,8 +135,8 @@ def validator(train_data: pd.DataFrame,
     predict_oof : bool
         Whether to return out of fold predictions on the logs
 
-    return_train_score : bool
-        Whether to include train scores
+    return_eval_fn_on_train : bool
+        Whether to apply eval_fn to the training sets and return the evaluation logs in the train logs
 
     return_all_train_logs : bool
         Whether to return the train logs corresponding to all the splits or to return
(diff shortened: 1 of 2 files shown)


src/fklearn/validation/validator.py
```diff
@@ -63,6 +67,8 @@ def validator_iteration(data: pd.DataFrame,
     warnings.warn(empty_set_warn) if train_data.shape[0] == 0 else None  # type: ignore
 
     predict_fn, train_out, train_log = train_fn(train_data)
+    if return_train_score:
+        train_log = assoc(train_log, "eval_results", eval_fn(train_out))
```
erickisos, line 71: Just a quick question, why do you choose `assoc` instead of this?
```suggestion
        train_log['eval_results'] = eval_fn(train_out)
```

erickisos (commented):
I know that this is still in progress, but in general, it looks good to me so I just left one question

src/fklearn/validation/validator.py
diegoarri91: Hi Eric! Thanks for your comments! The `assoc` function is used multiple times in this file, everytime a new key-value pair is added to a dictionary, so I used it to be consistent. I guess it's more functional style than assigning the value to the new key because instead of mutating the already existing dictionary the `assoc` function returns a new dictionary with the new key-value pair added to it.

erickisos: That's a great reason, consistency is always important, thanks for your answer Diego!

src/fklearn/validation/validator.py
```diff
@@ -161,26 +176,31 @@ def get_perturbed_columns(perturbator: PerturbFnType) -> List[str]:
         return args['cols'] if args else []
 
     train_logs, validator_logs = zip(*map(_join_split_log, zipped_logs))
-    first_train_log = first(train_logs)
+    if return_all_train_logs:
+        train_logs = {"train_log": [log['train_log'] for log in train_logs]}
```
robotenique, line 180:
```suggestion
        train_logs = {"train_log": [log["train_log"] for log in train_logs]}
```

robotenique (changes requested):
Thanks for the contribution, leaving some general comments!

src/fklearn/validation/validator.py
```diff
@@ -222,6 +243,9 @@ def parallel_validator(train_data: pd.DataFrame,
     predict_oof : bool
         Whether to return out of fold predictions on the logs
 
+    return_train_score : bool
```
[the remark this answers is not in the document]
diegoarri91: This is because the **current** behavior differs for both functions. For some reason with current behavior, `parallel_validator` always returns only the first train log while `parallel_validator` always returns all train logs. I feel they could be made consistent but in this PR I opted to preserve this behavior for both (when using default parameters).

src/fklearn/validation/validator.py
```diff
@@ -51,6 +52,9 @@ def validator_iteration(data: pd.DataFrame,
...
+        Whether to include train scores
```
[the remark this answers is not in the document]
diegoarri91: You are right. I am applying eval_fn to the training set and returning the result of this in the train logs. I renamed the parameter and changed the descriptions. Thanks!!

src/fklearn/validation/validator.py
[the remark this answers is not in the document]
robotenique: got it

src/fklearn/validation/validator.py
```diff
@@ -21,6 +21,7 @@ def validator_iteration(data: pd.DataFrame,
                         train_fn: LearnerFnType,
                         eval_fn: EvalFnType,
                         predict_oof: bool = False,
+                        return_eval_fn_on_train: bool = False,
```
robotenique, line 24: just a slight nitpick, I'd name it `return_eval_logs_on_train`,  since what you're returing is the logs of the evaluation and not the evaluation function

robotenique (approved):
👏

robotenique (approved):

PR #197, opened 2022-05-20
Add lgbm categorical features support

### LINKED ISSUE #194: Categorical Features in lgbm_classification_learner not working
I was using lgbm_classification_learner with categorical features and currently it's not possible to deal with them. It's possible to pass "categorical_feature" via "extra_params" but an error is raised because lgbm_classification_learner uses df.values (numpy array) instead of the original pandas dataframe. 

To avoid this behavior, I remove the .values and it worked

Why is it using "df.values"? Because of the performance?

If you agree with the solution of removing ".values" I could do a PR.

fberanizo:
### Status
**READY**

### Todo list
- [ ] Documentation
- [ ] Tests added and passed
- [ ] Issue: closes #194

### Background context

Older versions of LigthGBM used to allow passing `categorical_feature`through the argument `param`, but recent versions raise the following warning:

```
UserWarning: categorical_feature keyword has been found in `params` and will be ignored.
Please use categorical_feature argument of the Dataset constructor to pass this parameter.
  .format(key))
```

It is dubious (for me) if the warning is misleading and the option still works. [A contributor at lightgbm said that the correct way to pass them is through the Dataset object](https://github.com/microsoft/LightGBM/issues/1408#issuecomment-530096617).

Adds a new option `categorical_features` to `lgbm_classification_learner`. It allows a list of column names that should be treated as categorical features. Further instructions are found in https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst

_(write the remaining problems or questions that this PR does not solve)_

### DIFF
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -502,6 +502,7 @@ def lgbm_classification_learner(df: pd.DataFrame,
                                 learning_rate: float = 0.1,
                                 num_estimators: int = 100,
                                 extra_params: LogType = None,
+                                categorical_features: List[str] = "auto",
                                 prediction_column: str = "prediction",
                                 weight_column: str = None,
                                 encode_extra_cols: bool = True) -> LearnerReturnType:
@@ -549,6 +550,11 @@ def lgbm_classification_learner(df: pd.DataFrame,
         https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst
         If not passed, the default will be used.
 
+    categorical_features : list of str, or 'auto', optional (default="auto")
+        A list of column names that should be treated as categorical features.
+        See the categorical_feature hyper-parameter in:
+        https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst
+
     prediction_column : str
         The name of the column with the predictions from the model.
 
@@ -570,7 +576,7 @@ def lgbm_classification_learner(df: pd.DataFrame,
     features = features if not encode_extra_cols else expand_features_encoded(df, features)
 
     dtrain = lgbm.Dataset(df[features].values, label=df[target], feature_name=list(map(str, features)), weight=weights,
-                          silent=True)
+                          silent=True, categorical_feature=categorical_features)
 
     bst = lgbm.train(params, dtrain, num_estimators)
 
--- a/tests/training/test_classification.py
+++ b/tests/training/test_classification.py
@@ -431,6 +431,7 @@ def test_lgbm_classification_learner():
                                                  learning_rate=0.1,
                                                  num_estimators=20,
                                                  extra_params={"max_depth": 4, "seed": 42},
+                                                 categorical_features=["x2"],
                                                  prediction_column="prediction",
                                                  weight_column="w")
 


src/fklearn/training/classification.py
```diff
@@ -570,7 +576,7 @@ def lgbm_classification_learner(df: pd.DataFrame,
     features = features if not encode_extra_cols else expand_features_encoded(df, features)
 
     dtrain = lgbm.Dataset(df[features].values, label=df[target], feature_name=list(map(str, features)), weight=weights,
```
jmoralez, line 578: `categorical_feature='auto'` means treating the columns that have categorical data type in a pandas dataframe as categorical for training. Since this takes the `.values` attribute it will be ignored. I propose removing the `.values` here.

tests/training/test_classification.py
```diff
@@ -431,6 +431,7 @@ def test_lgbm_classification_learner():
                                                  learning_rate=0.1,
                                                  num_estimators=20,
                                                  extra_params={"max_depth": 4, "seed": 42},
+                                                 categorical_features=["x2"],
```
jmoralez, line 434: We should probably test that `x2` is actually being treated as categorical. We can check the split types of the booster object, because for categorical feature the decision type is '==' instead of '<='

jmoralez (changes requested):

tests/training/test_classification.py
fberanizo: Thank you for the feedback! 

I added a new test case and a check similar to this:
`assert any(tree['num_cat'] > 0 for tree in log['object'].dump_model()['tree_info'])`

Please let me know if any other changes are necessary.

jmoralez (dismissed):

fberanizo:
I'm worried that the change below might alter the behavior of existing models (due to the use of pandas + categorical_feature='auto' setting).<br>
Is this still a reasonable change?

```diff
-  dtrain = lgbm.Dataset(df[features].values, label=df[target], feature_name=list(map(str, features)), weight=weights,
-                          silent=True)
+    dtrain = lgbm.Dataset(df[features], label=df[target], feature_name=list(map(str, features)), weight=weights,
+                          silent=True, categorical_feature=categorical_features)
```

fberanizo:
I'm worried that the change below might alter the behavior of existing models (due to the use of pandas + categorical_feature='auto' setting).<br>
Is this still a reasonable change?

```diff
-  dtrain = lgbm.Dataset(df[features].values, label=df[target], feature_name=list(map(str, features)), weight=weights,
-                          silent=True)
+    dtrain = lgbm.Dataset(df[features], label=df[target], feature_name=list(map(str, features)), weight=weights,
+                          silent=True, categorical_feature=categorical_features)
```

jmoralez:
I'd say this change is for the better because it's the same behavior as using LightGBM directly. However, taking a closer look at the code I see that when predicting the `values` attribute is used as well: https://github.com/nubank/fklearn/blob/a3de09a6949089f47633e64b536a808650d7e9d1/src/fklearn/training/regression.py#L487
https://github.com/nubank/fklearn/blob/a3de09a6949089f47633e64b536a808650d7e9d1/src/fklearn/training/classification.py#L70
So it'll definitely cause some headaches if we don't change it there as well. Yet for SHAP the dataframe is used:
https://github.com/nubank/fklearn/blob/a3de09a6949089f47633e64b536a808650d7e9d1/src/fklearn/training/regression.py#L492
I think it'd be best to use the dataframe everywhere to not cause any surprises on the user, and using `values` isn't always more efficient than the dataframe. Also the dataframe allows to use the categorical features in their "raw" form, i.e. if we leave the `.values` there the user will always have to convert them to integer codes.

tiagorm (dismissed):

tiagorm:
> I'm worried that the change below might alter the behavior of existing models (due to the use of pandas + categorical_feature='auto' setting). Is this still a reasonable change?
> 
> ```diff
> -  dtrain = lgbm.Dataset(df[features].values, label=df[target], feature_name=list(map(str, features)), weight=weights,
> -                          silent=True)
> +    dtrain = lgbm.Dataset(df[features], label=df[target], feature_name=list(map(str, features)), weight=weights,
> +                          silent=True, categorical_feature=categorical_features)
> ```

Agree with the above comments, this change itself looks good but is better to review the `.values` usages

fberanizo:
Hi guys!
Sorry for the (very) late response. 
Recently, @isphus1973 and @hellenlima reached out to me asking about this feature, and I finally spared some time to finish it.

The recent commits:
- Added the same changes to `lgbm_regression_learner`
- Removed occurrences of `df.values`, as you suggested

isphus1973 (approved):
Fantastic

PR #201, opened 2022-07-07
Causal S-Learner

GiulioCMSanto:
### Status
**IN DEVELOPMENT**

### Todo list
- [x] Documentation
- [x] Lint
- [x] Tests added and passed
- [x] Example Notebook
- [x] Docstrings

### Background context
A very interesting and useful type of causal model are uplift models, 
where we want to identify how each sample responds to a given treatment,
and what the effect of that treatment is. That is, we want to know what 
is the probability of a sample converting, given that it received a nudge.
Meta-learns are examples of causal models, in which the CATE represents how 
each unit will respond to a given treatment [1]. In addition, these models 
have the advantage of using conventional machine learning models.

Therefore, the main idea of this pull request is to bring an implementation 
to S-learner [2], following the functional pattern of fklearn and, more than 
that, allowing it to be integrated with the other fklearn functionalities, such
as pipelines.

References:

[1] https://matheusfacure.github.io/python-causality-handbook/21-Meta-Learners.html

[2] https://causalml.readthedocs.io/en/latest/methodology.html

So far, a new `meta_learners.py` script is being included in the `causal/cate_learning`
package.

A good start is reviewing the `causal_s_classification_learner` function code and,
as you ready, go through the auxiliary functions.

Some remaining topics to think about:

- The way the code is implemented is already almost extrapolable 
to regression models. My idea is to think about this a bit more and maybe 
leave the classifier as a generic model that can be used for regression 
and/or classification;

- Discuss about the log format


src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,233 @@
...
+    return control_pred_df[prediction_column].values
+
+
+def _predict_for_treatment(
+    df: pd.DataFrame, learner_fcn: PredictFnType, prediction_column: str
+) -> np.array:
+    treatment_flag = np.ones(df.shape[0])
+    df[TREATMENT_FEATURE] = treatment_flag
+    treatment_pred_df = learner_fcn(df)
+    return treatment_pred_df[prediction_column].values
```
fabiofabris, line 93: These are very similar. I think it would be better to have a _predict_for_treatment_value that would receive the value (0 or 1) as an argument.

fabiofabris (changes requested):

src/fklearn/causal/cate_learning/meta_learners.py
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,231 @@
...
+        )
+
+        uplift_cols.append(f"treatment_{treatment}__uplift")
+        scored_df[uplift_cols[-1]] = (
+            scored_df[f"treatment_{treatment}__{prediction_column}_on_treatment"]
+            - scored_df[f"treatment_{treatment}__{prediction_column}_on_control"]
+        )
+
+    scored_df["uplift"] = scored_df[uplift_cols].max(axis=1).values
+    scored_df["suggested_treatment"] = scored_df[uplift_cols].idxmax(axis=1).values
```
nicolas-behar, line 128: we should also recommend not doing anything in the case the uplift is negative

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,231 @@
...
+    scored_df.drop(columns=[TREATMENT_FEATURE], inplace=True)
+
+    return scored_df
+
+
+@curry
+def causal_s_classification_learner(
+    df: pd.DataFrame,
+    treatment_col: str,
+    control_name: str,
```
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
GiulioCMSanto: ✅

robotenique:
Have you ran the formatting?

```python -m pip install -q flake8
python -m flake8 \
  --ignore=E731,W503 \
  --filename=\*.py \
  --exclude=__init__.py \
  --show-source \
  --statistics \
  --max-line-length=120 \
  src/ tests/
```

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,271 @@
...
+    df = pd.DataFrame(
+        {
+            "x1": [1.3, 1.0, 1.8, -0.1],
+            "x2": [10, 4, 15, 6],
+            TREATMENT_FEATURE: [1.0, 1.0, 0.0, 0.0],
+            "target": [1, 1, 1, 0],
+        }
+    )
+
+    learner_binary = logistic_classification_learner(
```
eduardodx, line 168: Should we add a random_state here, since we are comparing specific values?

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,271 @@
...
+                "treatment_A_",
+                "treatment_A_",
+            ],
+        }
+    )
+
+    treatments = ["A", "B"]
+    control_name = "control"
+
+    mock_learner = create_autospec(logistic_classification_learner)
```
eduardodx, line 249: Could you please add some comments to this section so we can better understand what's going on?

eduardodx (commented):
Well done! Very nice :D

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,271 @@
...
+        treatment_col="treatment",
+        control_name="control",
+        treatments=treatments,
+    )
+
+    assert len(learners) == len(treatments)
+    assert len(logs) == len(treatments)
+
+
+def test__predict_by_treatment_flag_positive():
```
GiulioCMSanto, line 158: @nicolas-behar, I think a better test here would be to Mock the learner... We just need to test if the function is working. We would mock the learner and build a fake learner that returns, for instance, the treatment flag... and check if the prediction (which will be the flag) matches the flag column.

tests/causal/cate_learning/test_meta_learners.py
GiulioCMSanto: This is just a mock...

GiulioCMSanto: This is just a mock...

tests/causal/cate_learning/test_meta_learners.py
GiulioCMSanto: I think we should mock the learner, instead. What do you think?

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,271 @@
...
+                "treatment_A_",
+                "treatment_A_",
+            ],
+        }
+    )
+
+    treatments = ["A", "B"]
+    control_name = "control"
+
+    mock_learner = create_autospec(logistic_classification_learner)
```
GiulioCMSanto, line 249: I'm not sure, but I think is better to mock the entire _predict_by_treatment_flag function, which has its own unit test, and make a side effect for its output that contemplates the two necessary cases: flag = True and flag = False. We should not be restricted to the particular logistic_classification_learner object...

GiulioCMSanto: you can just make a @patch for it...

nicolas-behar: Thanks!
Just simplified the test by creating a dummy model that returns the treatment column, which will always be 1 or 0. so I could test the local variable in the function.

tests/causal/cate_learning/test_meta_learners.py
nicolas-behar: Thanks!
Just simplified the test by creating a dummy model that returns the treatment column, which will always be 1 or 0. so I could test the local variable in the function.

tests/causal/cate_learning/test_meta_learners.py
nicolas-behar: I removed the values and now we are testing the input with 1 and 0 values only, since we don't need to test the learner.

tests/causal/cate_learning/test_meta_learners.py
GiulioCMSanto: @nicolas-behar

GiulioCMSanto: @nicolas-behar

tests/causal/cate_learning/test_meta_learners.py
fabiofabris: I agree, and the change should be minimal.

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,394 @@
...
+                "control",
+                "control",
+                "B",
+                "control",
+            ],
+            "target": [1, 1, 1, 0, 0, 1, 0, 0, 1],
+        }
+    )
+
+    mock_model = create_autospec(logistic_classification_learner)
```
fabiofabris, line 374: ditto.

fabiofabris (changes requested):

tests/causal/cate_learning/test_meta_learners.py
nicolas-behar: included a brief explanation on why there are 4 outputs to this test
should I add anything else?

tests/causal/cate_learning/test_meta_learners.py
nicolas-behar: patching _predict_by_treatment_flag is possible but we need to patch other functions too
keeping a mock learner looks more simple, wdyt?

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,394 @@
...
+    learners, logs = _fit_by_treatment(
+        df,
+        learner=learner_binary,
+        treatment_col="treatment",
+        control_name="control",
+        treatments=treatments,
+    )
+
+    assert len(learners) == len(treatments)
+    assert len(logs) == len(treatments)
```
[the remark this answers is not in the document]
nicolas-behar: done!

tests/causal/cate_learning/test_meta_learners.py
nicolas-behar: changed to mock_learner

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,394 @@
...
+
+    learners, logs = _fit_by_treatment(
+        df,
+        learner=learner_binary,
+        treatment_col="treatment",
+        control_name="control",
+        treatments=treatments,
+    )
+
+    assert len(learners) == len(treatments)
```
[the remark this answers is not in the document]
nicolas-behar: Done!

fabiofabris (approved):

fabiofabris (approved):

src/fklearn/types/types.py
```diff
@@ -18,6 +18,7 @@
 
 UncurriedLearnerFnType = Callable[..., LearnerReturnType]
 LearnerFnType = Callable[[pd.DataFrame], LearnerReturnType]
+LearnerMutableParametersFnType = Callable[..., LearnerFnType]
```
prcastro, line 21: Why do we need a new type here? It's not very clear to me

tests/causal/cate_learning/test_meta_learners.py
```diff
@@ -0,0 +1,402 @@
...
+    _create_treatment_flag,
+    _fit_by_treatment,
+    _predict_by_treatment_flag,
+    _simulate_treatment_effect,
+    causal_s_classification_learner,
+)
+
+from unittest.mock import create_autospec, patch
+
+from fklearn.types import LearnerFnType
```
prcastro, line 26: 💅 order the imports

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,268 @@
...
+        scored_df = _simulate_treatment_effect(
+            df=new_df,
+            treatments=unique_treatments,
+            learners=fitted_learners,
+            control_name=control_name,
+            prediction_column=prediction_column,
+        )
+        return scored_df
+
+    p.__doc__ = learner_pred_fn_docstring("causal_s_classification_learner", shap=True)
```
prcastro, line 256: what shap=True means here?

src/fklearn/types/types.py
GiulioCMSanto: Hey @prcastro 😄 I created this new type exclusively because the s-learner requires the creation of a new feature (T), that determines the treatment case (T=1) and the control case (T=0). Since we are creating this function upon fklearn, when we update the features from the classifier's partial function, mypy breaks, because the partial function only expects a dataframe as its input (it does not expect that you change the features parameter that were already provided). The solution I found was to create a more "flexible" type. But maybe there is a better way!

Here is where mypy breaks: https://github.com/nubank/fklearn/blob/a42339807e18a8cbdac5dc4fd8c568c01715ab5c/src/fklearn/causal/cate_learning/meta_learners.py#L229-L233

prcastro: The typing system in python don’t play well with partial function, that’s a fact. I’d do one of two things:

- Just declare the learner as `Callable` without specifying inputs and outputs
- Rename the type `PartialLearnerFnType` so it’s more clear

I prefer the first one.

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,268 @@
...
+def _predict_by_treatment_flag(
+    df: pd.DataFrame,
+    learner_fcn: PredictFnType,
+    is_treatment: bool,
+    prediction_column: str,
+) -> np.ndarray:
+    if is_treatment:
+        treatment_flag = np.ones(df.shape[0])
+    else:
+        treatment_flag = np.zeros(df.shape[0])
```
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,268 @@
...
+        The name of the column in `df` which contains the names of
+        the treatments or control to which each data sample was subjected.
+
+    control_name: str
+        The name of the control group.
+
+    prediction_column : str
+        The name of the column with the predictions from the provided learner.
+
+    learner: LearnerFnType
```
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
GiulioCMSanto: ✅

GiulioCMSanto: (removed)

src/fklearn/causal/cate_learning/meta_learners.py
```diff
@@ -0,0 +1,268 @@
...
+            prediction_column=prediction_column,
+        )
+        return scored_df
+
+    p.__doc__ = learner_pred_fn_docstring("causal_s_classification_learner", shap=True)
+
+    partial_log = {"causal_features": features_with_treatment}
+    partial_log.update(learners_logs)
+
+    log = {"causal_s_classification_learner": partial_log}
```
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/types/types.py
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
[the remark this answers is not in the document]
GiulioCMSanto: ✅

src/fklearn/causal/cate_learning/meta_learners.py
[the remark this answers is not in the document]
GiulioCMSanto: (sorry, didn't commit because I applied flake)

tests/causal/cate_learning/test_meta_learners.py
GiulioCMSanto: ✅

prcastro (approved):

PR #207, opened 2022-08-10
Vectorize apply_replacement

matheusfacure:
### Status
**IN DEVELOPMENT**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Pandas .apply is incredibly slow to run. Since this function is used in multiple learners, speeding it up should yield tremendous grains in performance. As a side note, **we should never introduce new call `.apply`**. They are a major source of headache 

Vectorize `apply_replacements `

### DIFF
--- a/src/fklearn/training/transformation.py
+++ b/src/fklearn/training/transformation.py
@@ -354,7 +354,7 @@ def apply_replacements(df: pd.DataFrame,
         Default value to replace when original value is not present in the `vec` dict for the feature
 
     """
-    def column_categorizer(col: str):
+    def column_categorizer(col: str) -> np.ndarray:
         replaced = df[col].map(vec[col])
         unseen = df[col].notnull() & replaced.isnull()
         replaced[unseen] = replace_unseen
--- a/src/fklearn/training/transformation.py
+++ b/src/fklearn/training/transformation.py
@@ -354,11 +354,8 @@ def apply_replacements(df: pd.DataFrame,
         Default value to replace when original value is not present in the `vec` dict for the feature
 
     """
-    def column_categorizer(col: str) -> np.ndarray:
-        replaced = df[col].map(vec[col])
-        unseen = df[col].notnull() & replaced.isnull()
-        replaced[unseen] = replace_unseen
-        return replaced
+    def column_categorizer(col: str) -> pd.Series:
+        return df[col].map(lambda x: vec[col].get(x, replace_unseen), na_action='ignore')
 
     categ_columns = {col: column_categorizer(col) for col in columns}
     return df.assign(**categ_columns)
--- a/src/fklearn/training/transformation.py
+++ b/src/fklearn/training/transformation.py
@@ -355,11 +355,10 @@ def apply_replacements(df: pd.DataFrame,
 
     """
     def column_categorizer(col: str):
-        return np.select(
-            [df[col].isna() | (df[col].dtype == "float"), ~df[col].isin(vec[col].keys())],
-            [np.nan, replace_unseen],
-            df[col].replace(vec[col])
-        ).astype(np.array(list(vec[col].values())).dtype)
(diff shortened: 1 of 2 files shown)


jmoralez:
WDYT about using map on the dicts instead?
```python
import numpy as np
import pandas as pd

nrows = 1_000_000
ncols = 20
n_unique_vals = 100
df = pd.DataFrame(
    np.random.randint(0, n_unique_vals, (nrows, ncols)),
    columns=[f'x{i}' for i in range(ncols)],
    dtype=str
)
cols_to_replace = [f'x{i}' for i in range(0, 20, 2)]
vec = {col: {str(i): str(i + 1) for i in range(n_unique_vals - 10)}  # last 10 values were not seen
       for col in cols_to_replace}  
# define one of the replacement columns as float
df['x0'] = df['x0'].astype('float')
vec['x0'] = {float(k): float(v) for k, v in vec['x0'].items()}
replace_unseen = -1

def apply_replacements(df, columns, vec, replace_unseen):
    def column_categorizer(col: str):
        return np.select(
            # the original had an and here so I guess it should be &
            [df[col].isna() & (df[col].dtype == "float"), ~df[col].isin(vec[col].keys())],
            [np.nan, replace_unseen],
            df[col].replace(vec[col])
        )
    return df.assign(**{col: column_categorizer(col) for col in columns})
%time res1 = apply_replacements(df, cols_to_replace, vec, replace_unseen)
# Wall time: 1min 22s

# proposal
def apply_replacements2(df, columns, vec, replace_unseen):
    def column_categorizer(col: str):
        replaced = df[col].map(vec[col])
        unseen = df[col].notnull() & replaced.isnull()
        replaced[unseen] = replace_unseen
        return replaced
    return df.assign(**{col: column_categorizer(col) for col in columns})
%time res2 = apply_replacements2(df, cols_to_replace, vec, replace_unseen)
# Wall time: 3.93 s

pd.testing.assert_frame_equal(res1, res2)
```

matheusfacure:
WOW! What is this magic? How does map works?

jmoralez:
The main difference is that replace only changes the values you provide in the dict, whereas map tries to replace all of them and when there isn't a match it sets the value to null, which in this case I think is helpful for us because we can get the ones that didn't match very easily.

src/fklearn/training/transformation.py
```diff
@@ -354,9 +354,12 @@ def apply_replacements(df: pd.DataFrame,
...
 
     """
-    column_categorizer = lambda col: df[col].apply(lambda x: (np.nan
-                                                              if isinstance(x, float) and np.isnan(x)
-                                                              else vec[col].get(x, replace_unseen)))
+    def column_categorizer(col: str) -> np.ndarray:
+        replaced = df[col].map(vec[col])
+        unseen = df[col].notnull() & replaced.isnull()
+        replaced[unseen] = replace_unseen
+        return replaced
```
denismr, line 361:
```suggestion
        return replaced.mask(unseen, replace_unseen)
```

src/fklearn/training/transformation.py
```diff
@@ -354,11 +354,8 @@ def apply_replacements(df: pd.DataFrame,
         Default value to replace when original value is not present in the `vec` dict for the feature
 
     """
-    def column_categorizer(col: str) -> np.ndarray:
-        replaced = df[col].map(vec[col])
-        unseen = df[col].notnull() & replaced.isnull()
-        replaced[unseen] = replace_unseen
-        return replaced
+    def column_categorizer(col: str) -> pd.Series:
+        return df[col].map(lambda x: vec[col].get(x, replace_unseen), na_action='ignore')
```
matheusfacure, line 358: is this faster than `.apply`? I don't know how map works under the hood, but if it implements a for loop in the backend for lambda function, than its just as bad as apply no?

PR #208, opened 2022-08-16
Bugfix for some cases where validator fails

nicolas-behar:
### Status
**IN DEVELOPMENT**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
Depending on the data - if it has gaps that makes all the training start and end date be NaT, the validator function will break.

This PR should fix the mentioned issue by skipping the folds where there is no training or testing data and saving these failure logs. The changes will only take effect if "drop_empty_folds" is set to be True, otherwise, the implementation of this function will remain as it was before.

### Related PR
- https://github.com/nubank/fklearn/pull/210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [2.1.1] - 2022-08-17
+- **Bug Fix**
+  - Fix validator behavior when receiving data containing gaps and a time based split function that could generate empty 
+    training and testing folds and then break. 
+    The argument `drop_empty_folds` can be set to `True` to drop invalid folds from validation and store them in the 
+    log.
+ 
 ## [2.1.0] - 2022-07-25
 - **Enhancement**
     - Add optional parameter `return_eval_logs_on_train` to the `validator` function,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,5 @@
 # Changelog
 
-## [2.1.1] - 2022-08-17
-- **Bug Fix**
-  - Fix validator behavior when receiving data containing gaps and a time based split function that could generate empty 
-    training and testing folds and then break. 
-    The argument `drop_empty_folds` can be set to `True` to drop invalid folds from validation and store them in the 
-    log.
- 
 ## [2.1.0] - 2022-07-25
 - **Enhancement**
     - Add optional parameter `return_eval_logs_on_train` to the `validator` function,
--- a/src/fklearn/validation/validator.py
+++ b/src/fklearn/validation/validator.py
@@ -178,8 +178,7 @@ def validator(train_data: pd.DataFrame,
                        map(fold_iter),
                        partial(zip, logs))
 
-    def clean_logs(log_tuple: typing.Tuple[zip[typing.Tuple[Any, Any]], typing.List[Any]]) -> \
-            typing.Tuple[zip[typing.Tuple[Any, Any]], typing.List[Any]]:
+    def clean_logs(log_tuple: typing.Tuple) -> (typing.Tuple, typing.List):
         split_log_error = list()
 
         new_validator_logs = list()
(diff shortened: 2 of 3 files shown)


src/fklearn/validation/validator.py
```diff
@@ -157,14 +164,36 @@ def validator(train_data: pd.DataFrame,
 
     def fold_iter(fold: Tuple[int, Tuple[pd.Index, pd.Index]]) -> LogType:
         (fold_num, (train_index, test_indexes)) = fold
-        return validator_iteration(train_data, train_index, test_indexes, fold_num,
-                                   train_fn, eval_fn, predict_oof, return_eval_logs_on_train, verbose)
+        if (len(train_index) <= 1 or len(test_indexes[0]) == 0) and drop_empty_folds:
```
GiulioCMSanto, line 167: The test_indexes is a list of indexes. A suggestion is the following:

```suggestion
        
        test_contains_null_folds = max(map(lambda x: len(x) == 0, test_indexes))
        train_fold_is_null = len(train_index) == 0

        if (train_fold_is_null or test_contains_null_folds) and drop_empty_folds:
```

tests/validation/test_validator.py
```diff
@@ -1,47 +1,57 @@
...
 
 perturb_fn_train = identity
-perturb_fn_test = perturbator(cols=['rows'], corruption_fn=nullify(perc=0.25))
+perturb_fn_test = perturbator(cols=["rows"], corruption_fn=nullify(perc=0.25))
 
 
-data = pd.DataFrame({
-    'rows': ['row1', 'row2', 'row3', 'row4']
-})
+data = pd.DataFrame({"rows": ["row1", "row2", "row3", "row4"]})
```
GiulioCMSanto, line 54: Maybe using a fixture here?

```suggestion
@pytest.fixture
def data():
    return pd.DataFrame({"rows": ["row1", "row2", "row3", "row4"]})
```

tests/validation/test_validator.py
```diff
@@ -1,47 +1,57 @@
...
 
 
 def test_validator_iteration():
```
GiulioCMSanto, line 57:
```suggestion
def test_validator_iteration(data):
```

tests/validation/test_validator.py
```diff
@@ -67,23 +79,136 @@ def test_validator_iteration():
...
+
+    validator_log = validator(
+        df_gap, split_fn_gap, model, eval_fn_gap, drop_empty_folds=True
+    )["validator_log"]
+
+    assert len(validator_log[0]) == 3
+    assert len(validator_log[0]["eval_results"]) == 1
 
 
 def test_parallel_validator():
```
GiulioCMSanto, line 214:
```suggestion
def test_parallel_validator(data):
```

tests/validation/test_validator.py
```diff
@@ -67,23 +79,136 @@ def test_validator_iteration():
 
 
 def test_validator():
```
GiulioCMSanto, line 81: Great! I would suggest one more test, which is when the splitters provides a list of indexes for the testing index, i.e., fold = [train_index, list of test_index]

GiulioCMSanto (changes requested):

GiulioCMSanto (commented):
I am getting an error in the following line when running the test_validator test:

https://github.com/nubank/fklearn/blob/bfdc44d40cb194907fc59f022ddd38daf7a2e7c9/src/fklearn/validation/validator.py#L211

It seems that the issue is the following:

Since you are using a lightgbm, in the zipped_log we have the training log which, in turns, have the lightgbm version (lightgbm._ _version _ _). However, lightgbm is not being imported, raising an exception.

vinipd (dismissed):
LGTM.

GiulioCMSanto (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [2.1.1] - 2022-08-17
```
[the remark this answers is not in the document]
nicolas-behar: I will open a new PR with this bump and changelog, and include the S-Learner info also, since it was not bumped yet.

GiulioCMSanto (approved):

CHANGELOG.md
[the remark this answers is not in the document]
nicolas-behar: done it at https://github.com/nubank/fklearn/pull/210

Issue #117, opened 2020-01-13
Allow fit params for LightGBM


ValeriaGomes:
### Describe the feature and the current state.
- LightGBM has several parameters inside the `fit` method. One possible solution is to add a `fit_params` into the LightGBM function in FkLearn for increased flexibility.

### Will this change a current behavior? How?
We would be able to configure the fit function using FkLearn.

### Additional Information
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier.fit

sadikneipp:
I think the extra params argument covers that, right?
[https://github.com/nubank/fklearn/blob/master/src/fklearn/training/classification.py#L507:L511](url)

sadikneipp:
Hmm I'm wrong, this argument only covers the hyperparameter section here https://lightgbm.readthedocs.io/en/latest/Parameters.html

sadikneipp:
Any specific behaviour you feel we are missing @ValeriaGomes? I see there's some overlap between what `extra_params` cover and the params in the link you sent

ValeriaGomes:
Good point!  I hadn't seen the `extra_params`. I'll check if there're parameters missing and if it's not the case, I'll close the issue.

robotenique:
🆙  on this, I think with the current learners we cannot pass a validation set to work with early stopping in LightGBM.

PR #211, opened 2022-09-02
Including LGBM parameters to lgbm_classification_learner

### LINKED ISSUE #194: Categorical Features in lgbm_classification_learner not working

nicolas-behar:
### Status
**IN DEVELOPMENT**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes #194, #117 

### Background context
Some of the already implemented LGBM parameters can't be used by the lgbm_classification_learner function.

We are including these parameters by changing the lgbm.train call function with these parameters, that can be passed by the user if needed.

### Related PRs
https://github.com/nubank/fklearn/pull/197

### DIFF
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -1,7 +1,9 @@
-from typing import List, Any
+from typing import List, Any, Optional, Callable, Tuple, Union
 
 import numpy as np
 import pandas as pd
+from lightgbm import Booster
+from pathlib import Path
 from toolz import curry, merge, assoc
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.linear_model import LogisticRegression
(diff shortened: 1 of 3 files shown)


src/fklearn/training/classification.py
```diff
@@ -557,6 +571,44 @@ def lgbm_classification_learner(df: pd.DataFrame,
 
     encode_extra_cols : bool (default: True)
         If True, treats all columns in `df` with name pattern fklearn_feat__col==val` as feature columns.
+
+    valid_sets :
```
eduardodx, line 575: Missing description.

src/fklearn/training/classification.py
```diff
@@ -557,6 +571,44 @@ def lgbm_classification_learner(df: pd.DataFrame,
...
+
+    valid_names :
```
eduardodx, line 577: Missing description.

eduardodx (changes requested):

fabiofabris (approved):

eduardodx (approved):
LGTM! 🚀

PR #213, opened 2022-09-20
Replaced default argument min_df from 20 to 0 on TFIDF

raphaeldayan-nubank:
On sklearn min_df defaults to 0, it should default to 0 as well in fklearn. This argument is not specified in the docstring and was messing up with the performance of my sentiment classifier. I lost 20 points of recall because of this argument and had to spent hours figuring out that this was the problem. Could we change that argument to be consistent with sklearn?

### DIFF
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -446,7 +446,7 @@ def nlp_logistic_classification_learner(df: pd.DataFrame,
     """
 
     # set default params
-    default_vect_params = {"strip_accents": "unicode", "min_df": 20}
+    default_vect_params = {"strip_accents": "unicode", "min_df": 0}
     merged_vect_params = default_vect_params if not vectorizer_params else merge(default_vect_params, vectorizer_params)
 
     default_clf_params = {"C": 0.1, "multi_class": "ovr", "solver": "liblinear"}
--- a/src/fklearn/training/classification.py
+++ b/src/fklearn/training/classification.py
@@ -446,7 +446,7 @@ def nlp_logistic_classification_learner(df: pd.DataFrame,
     """
 
     # set default params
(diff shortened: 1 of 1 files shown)


hellenlima:
I see two options here:
1. Keep the default from scikit-learn (as @raphaeldayan-nubank suggested)
2. Make the [documentation](https://fklearn.readthedocs.io/en/latest/api/fklearn.training.html#fklearn.training.classification.nlp_logistic_classification_learner) **clear** about this difference

I prefer the first option, because it is more intuitive.

src/fklearn/training/classification.py
```diff
@@ -446,7 +446,7 @@ def nlp_logistic_classification_learner(df: pd.DataFrame,
     """
 
     # set default params
-    default_vect_params = {"strip_accents": "unicode", "min_df": 20}
+    default_vect_params = {"strip_accents": "unicode", "min_df": 0}
```
hellenlima, line 449: [scikit-learn default]( https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html) is actually min_df=1.
```suggestion
    default_vect_params = {"strip_accents": "unicode", "min_df": 1}
```

hellenlima (changes requested):

src/fklearn/training/classification.py
raphaeldayan-nubank: you are right, it's max_df=1, thank you!

hellenlima (approved):

gcbeltramini (approved):
Ref.: https://github.com/scikit-learn/scikit-learn/blob/7c2a58d51f4528827e9bfe9c43d06c5c1716bfb8/sklearn/feature_extraction/text.py#L1856

PR #217, opened 2022-11-28
Bumping the range of scikit-learn supported versions

peguerosdc:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

(Nothing had to change in here)

### Background context
Requiring a version of `scikit-learn` strictly lower tan `0.24` prevents the installation of `fklearn` in python 3.9. More context in [this Slack thread](https://nubank.slack.com/archives/CFCE6NJKS/p1669098405078879).

Versions of `scikit-learn` that can be installed in python 3.9 (and don't include any major change) are supported. The only breaking change is in `IsolationForest`, but a version check is added for backwards compatibility.

Check the bump in the range of the `scikit-learn` supported versions and, if interested, check about the `behaviour` param in the [IsolationForest's doc](https://scikit-learn.org/0.22/modules/generated/sklearn.ensemble.IsolationForest.html?highlight=isolationforest).


requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib>=0.13.2,<2
 numpy>=1.16.4,<2
 pandas>=0.24.1,<2
-scikit-learn>=0.21.2,<0.24.0
+scikit-learn>=0.21.2,<=0.24.2
```
gcbeltramini, line 4:
```suggestion
scikit-learn>=0.21.2,<0.25
```

gcbeltramini:
We need to figure out why the test is not running.

peguerosdc:
I think CircleCI needs to be set up again

Screen Shot 2022-12-05 at 9 43 32

fpingas:
Implemented in https://github.com/nubank/fklearn/pull/227

PR #220, opened 2022-12-20
Add ascending parameter causal validation

### LINKED ISSUE #204: Add an option to order by ascending/descending prediction in cumulative effect curves
### Instructions
- Delete everything between parenthesis _(...)_
- Remove the sections that are not relevant

In the causal validation module, it would be useful to add an `ascending` parameter for the cumulative effect and cumulative gain curves.

The current state is to [order predictions descending](https://github.com/nubank/fklearn/blob/ac4a3ee958392a9de7c6f941cb75997559432c94/src/fklearn/causal/validation/curves.py#L98):
```
ordered_df = df.sort_values(prediction, ascending=False).reset_index(drop=True)
```

If we add an `ascending: bool = False` argument to the `cumulative_effect_curve`, `cumulative_gain_curve`, `relative_cumulative_gain_curve`, and `effect_curves`, a user could modify how these effects are computed, whether to do them ascending or descending by the prediction column.

Not if the user does not explicitly change the argument to `ascending=False`.

### Additional Information

The new definition of `cumulative_effect_curve` would look like this:

```
@curry
def cumulative_effect_curve(df: pd.DataFrame,
                            treatment: str,
                            outcome: str,
                            prediction: str,
                            min_rows: int = 30,
                            steps: int = 100,
                            effect_fn: EffectFnType = linear_effect,
                            ascending: bool = False) -> np.ndarray:
    """
    Orders the dataset by prediction and computes the cumulative effect curve according to that ordering

    Parameters
    ----------
    df : Pandas' DataFrame
        A Pandas' DataFrame with target and prediction scores.

    treatment : Strings
        The name of the treatment column in `df`.

    outcome : Strings
        The name of the outcome column in `df`.

    prediction : Strings
        The name of the prediction column in `df`.

    min_rows : Integer
        Minimum number of observations needed to have a valid result.

    steps : Integer
        The number of cumulative steps to iterate when accumulating the effect

    effect_fn : function (df: pandas.DataFrame, treatment: str, outcome: str) -> int or Array of int
        A function that computes the treatment effect given a dataframe, the name of the treatment column and the name
        of the outcome column.

    ascending : bool
        Whether the prediction column should be ordered ascending or not. Default is False.


    Returns
    ----------
    cumulative effect curve: Numpy's Array
        The cumulative treatment effect according to the predictions ordering.
    """

    size = df.shape[0]
    ordered_df = df.sort_values(prediction, ascending=ascending).reset_index(drop=True)
    n_rows = list(range(min_rows, size, size // steps)) + [size]
    return np.array([effect_fn(ordered_df.head(rows), treatment, outcome) for rows in n_rows])
```

MarianaBlaz:
### Instructions
- Follow the instructions in [README.md](../blob/master/README.md)
- Delete everything between parenthesis _(...)_
- Remove the sections that are not relevant

### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed
- [x] Issue: closes #204 

### Background context
In the causal validation module and the [curves file](https://github.com/nubank/fklearn/blob/master/src/fklearn/causal/validation/curves.py), it would be useful to add an ascending parameter for the cumulative effect and cumulative gain curves.

The current state is to [order predictions descending](https://github.com/nubank/fklearn/blob/ac4a3ee958392a9de7c6f941cb75997559432c94/src/fklearn/causal/validation/curves.py#L98):

ordered_df = df.sort_values(prediction, ascending=False).reset_index(drop=True)
If we add an ascending: bool = False argument to the cumulative_effect_curve, cumulative_gain_curve, relative_cumulative_gain_curve, and effect_curves, a user could modify how these effects are computed, whether to do them ascending or descending by the prediction column.

A model could output a prediction that is not necessarily positively related to the effect to be computed, so adding an option to order this relationship differently allows for effects and gains with negatively related predictions and outcomes to be computed adequately.

The changes are applied to curves.py and also on auc.py on the causal-effect module.

Reviewing causal-effect/curves as there are the definition of the functions from which all ordering behavior is propagated.

### DIFF
--- a/src/fklearn/causal/validation/auc.py
+++ b/src/fklearn/causal/validation/auc.py
@@ -13,6 +13,7 @@ def area_under_the_cumulative_effect_curve(df: pd.DataFrame,
                                            prediction: str,
                                            min_rows: int = 30,
                                            steps: int = 100,
+                                           ascending: bool = False,
                                            effect_fn: EffectFnType = linear_effect) -> float:
     """
      Orders the dataset by prediction and computes the area under the cumulative effect curve, according to that
@@ -38,6 +39,9 @@ def area_under_the_cumulative_effect_curve(df: pd.DataFrame,
      steps : Integer
          The number of cumulative steps to iterate when accumulating the effect
 
+    ascending : Boolean
+        Indicates if the dataset should be ordered ascending with respect to the prediction column
+
      effect_fn : function (df: pandas.DataFrame, treatment: str, outcome: str) -> int or Array of int
          A function that computes the treatment effect given a dataframe, the name of the treatment column and the name
          of the outcome column.
@@ -55,7 +59,7 @@ def area_under_the_cumulative_effect_curve(df: pd.DataFrame,
     step_sizes = [min_rows] + [t - s for s, t in zip(n_rows, n_rows[1:])]
 
     cum_effect = cumulative_effect_curve(df=df, treatment=treatment, outcome=outcome, prediction=prediction,
-                                         min_rows=min_rows, steps=steps, effect_fn=effect_fn)
+                                         min_rows=min_rows, steps=steps, ascending=ascending, effect_fn=effect_fn)
 
     return abs(sum([(effect - ate) * (step_size / size) for effect, step_size in zip(cum_effect, step_sizes)]))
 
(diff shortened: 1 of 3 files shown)


raphael-bruce (approved):
LGTM

PR #233, opened 2023-08-15
Remove the scikit learn restriction and bump minimal python version to 3.8

otaviocv:
### Status
**READY**

### Todo list
- [x] Documentation
- [x] Tests added and passed

### Background context
fklearn had very old dependencies limiting the latest available features for many packages. With the increasing usage of M1 Macs working with these dependencies became an anoying issue. To overcome these and update fklearn we are making a major bump with all major packages bumped to their latest compatible version.

1. I wasn't able to bump the coverage packages.
2. The hashing function changes behavior from 3.8 to 3.9. This makes a conditional test necessary for hash value tests.


src/fklearn/resources/VERSION
```diff
@@ -1 +1 @@
-2.3.1
+2.4.0
```
laoikeda, line 1: Since removing support to python 3.6 is a breaking change, can you change this version to a major bump?
```suggestion
3.0.0
```

laoikeda (commented):
Thanks for the contribution! Can you check why the tests are failing and write a description in the PR?

src/fklearn/resources/VERSION
fpingas, line 1: Removing support for 3.6 has to be a major bump.

fpingas (changes requested):

requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib>=0.13.2,<2
 numpy>=1.16.4,<2
 pandas>=0.24.1,<2
```
gcbeltramini, line 3: Since this will be a breaking change in `fklearn`, why not bump `pandas`?

```suggestion
pandas>=0.24.1,<3
```

otaviocv: Given the opportunity, could we also remove the xgboost upper limitation on 1.5? We are now at 1.7.2

gcbeltramini: Yes. And if you find other dependencies that could be bumped, we could do it now. Let's look at the changes in these versions:

1. https://github.com/dmlc/xgboost/releases/tag/v1.5.0:

    > Breaking changes
    > - n_gpus was deprecated in 1.0 release and is now removed.
    > - Feature grouping in CPU hist tree method is removed, which was disabled long
ago. (https://github.com/dmlc/xgboost/pull/7018)
    > - C API for Quantile DMatrix is changed to be consistent with the new external memory
implementation. (https://github.com/dmlc/xgboost/pull/7082)

2. https://github.com/dmlc/xgboost/releases/tag/v1.6.0:

    > Python 3.7 is required as the lowest Python version. (https://github.com/dmlc/xgboost/pull/7682)

3. https://github.com/dmlc/xgboost/releases/tag/v1.7.0:

    > Breaking Changes
    > - The grow_local_histmaker updater is removed. This updater is rarely used in practice and has no test. We decided to remove it and focus have XGBoot focus on other more efficient algorithms. (https://github.com/dmlc/xgboost/pull/7992, https://github.com/dmlc/xgboost/pull/8091)
    > - Single precision histogram is removed due to its lack of accuracy caused by significant floating point error. In some cases the error can be difficult to detect due to log-scale operations, which makes the parameter dangerous to use. (https://github.com/dmlc/xgboost/pull/7892, https://github.com/dmlc/xgboost/pull/7828)
    > - Deprecated CUDA architectures are no longer supported in the release binaries. (https://github.com/dmlc/xgboost/pull/7774)
    > - As part of the federated learning development, the rabit module is replaced with the new collective module. It's a drop-in replacement with added runtime backend selection, see the federated learning section for more details (https://github.com/dmlc/xgboost/pull/8257)
    >
    > Python 3.8 is now the minimum required Python version. (https://github.com/dmlc/xgboost/pull/8071)

Based on how `xgboost` is used inside `fklearn` ([here](https://github.com/nubank/fklearn/blob/8fe082356cd0c920b97b0d1cf150af0d9417d978/src/fklearn/training/regression.py) and [here](https://github.com/nubank/fklearn/blob/8fe082356cd0c920b97b0d1cf150af0d9417d978/src/fklearn/training/classification.py)), the only important change is the minimum required Python version. Since official support to Python 3.7 is finished, we could also require at least Python 3.8.

gcbeltramini: I looked now at the changes in `pandas` 2.0.0 [here](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html), and didn't find anything that could affect `fklearn`. The possible change that I saw is in `value_counts`, but I tested how it is used in `fklearn`, and the behavior is the same.

otaviocv: I tried to bump libs as much as possible. Using pandas 2 forced xgboost to be bump to version 2 also. I am heavily relying on tests to assert that the current behavior is preserved.

laoikeda (approved):
LGTM

fpingas (approved):

